PR #16, opened 2020-12-03
Add release instructions

anuraaga:
*Description of changes:*  Adds RELEASING.md


RELEASING.md
```diff
@@ -0,0 +1,94 @@
...
+
+This triggers the release process, which builds the artifacts, updates the README with the new
+version numbers, commits the change to the README, publishes the artifacts, creates and pushes
+a git tag with the version number, and drafts a release with the agent artifact attached.
+
+## Announcement
+   
+Once the GitHub workflow completes, go to Github [release
+page](https://github.com/aws-observability/aws-java-otel-instrumentation/releases), and find the
+generated draft release to write release notes about the new release and add release notes.
```
willarmiros, line 18: Repetitive wording about release notes?

.github/workflows/patch-release-build.yml
```diff
@@ -0,0 +1,121 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      release-branch-name: ${{ steps.parse-release-branch.outputs.release-branch-name }}
+    steps:
+      - id: parse-release-branch
+        name: Parse release branch name
+        run: |
+          # Sets the release-branch-name output to the version number with the last non-period element replaced with an 'x' and preprended with v.
+          echo "::set-output name=release-branch-name::$(echo '${{ github.event.inputs.version }}' | sed -E 's/([^.]+)\.([^.]+)\.([^.]+)/v\1.\2.x/')"
+          # Sets the release-tag-name output to the version number with the last non-period element replace with a '0' and prepended with v
```
willarmiros, line 24: Why is the tag of the patch version being replaced with 0 for the tag? Just want to make sure this workflow isn't actually pushing a new tag, it's just using this tag to checkout the minor version release that we're patching.

.github/workflows/patch-release-build.yml
```diff
@@ -0,0 +1,121 @@
...
+      - name: Build release with Gradle
+        uses: burrunan/gradle-cache-action@v1
+        with:
+          arguments: build jibDockerBuild final -Prelease.version=${{ github.event.inputs.version }} --stacktrace
+        env:
+          PUBLISH_USERNAME: ${{ github.actor }}
+          PUBLISH_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
+          GRGIT_USER: ${{ github.actor }}
+          GRGIT_PASS: ${{ secrets.GITHUB_TOKEN }}
+          AWS_REGISTERY_ACCOUNT: ${{ secrets.AWS_ACCOUNT }}
```
willarmiros, line 79: Nit: typo in registry

RELEASING.md
```diff
@@ -0,0 +1,94 @@
...
+
+You can use `git log upstream/v$MAJOR.$((MINOR-1)).x..upstream/v$MAJOR.$MINOR.x --graph --first-parent`
+or the Github [compare tool](https://github.com/open-telemetry/opentelemetry-java/compare/)
+to view a summary of all commits since last release as a reference.
+
+## Patch Release
```
willarmiros, line 24: I'm confused why all this cherry-picking is necessary for a patch release and why it's not the same as any other release. If we're deciding to make a release, why can't we just input the new version with only the patch version incremented and use the original workflow?

willarmiros (commented):
A few questions/nits

.github/workflows/patch-release-build.yml
anuraaga: Yup exactly, this is only to compute the tag to checkout to create the base of the release (hotfix) branch. The version will be something like `1.2.1` and we need to make sure we checkout `1.2.0` before cherrypicking and push that to `1.2.x`

RELEASING.md
anuraaga: We might have done a lot of development on HEAD since the previous minor release. If we find a critical bug in a minor release within a day or something, we'd probably skip it and just release HEAD directly. But if we find it in a couple of weeks, we don't want to release new features, or significant potentially not-well-baked changes, in a patch release and should cherry-pick into a branch from the previous minor release tag to make sure they're not included. Depending on the nature of a hotfix and timeline, we could make a call on whether to just go straight to the next minor from HEAD, or release a patch version of the previous.

.github/workflows/patch-release-build.yml
willarmiros: Makes sense. So we are still pushing a tag for the patch itself later right (L107)? One edge case is if we're doing a second patch release, say `1.2.2`, shouldn't we checkout the `1.2.1` tag not the `1.2.0` tag then? So shouldn't it compute `MAJOR.MINOR.(PATCH-1)`  as the tag to start from?

RELEASING.md
willarmiros: Ok I see, makes sense to have that flexibility. Thanks for explaining!

willarmiros (approved):
LGTM - just the one lingering question about tags.

.github/workflows/patch-release-build.yml
anuraaga: After we've created the release branch, `1.2.x`, it's HEAD is always the previous patch release. An alternative would be to have branches for all of `1.2.1`, `1.2.2`, and then indeed we would check out from the previous tag - but it ends up with lots of branches so using `.x` is pretty conventional. It means we don't have to worry about checking out after the first patch release, we already have the release branch set up. It's why this is the `prepare-release-branch` job, it only runs once for the first patch (controlled by https://github.com/aws-observability/aws-otel-java-instrumentation/pull/16/files/e800882e7360e5b63ef23c24206528123a1c3526#diff-465a53fcd1e54767a6394a0fad650e8fd4872443de4dac07339011173c6c7480R34)

anuraaga: Added comments :)

willarmiros: Got ittttt so we only check out the minor release tag on the first patch release to make the `.x` branch then use that branch subsequently. Thanks for adding comments.

PR #17, opened 2020-12-03
Merge integration test flow into main build.

anuraaga:
*Description of changes:*

There seems to be an ordering issue with building integration test images - the otelagent base image may not be pushed before the apps are built causing them to use the old version. This merges everything into one, with Gradle dependencies, for better visbility of ordering (and some improved performance I guess).


willarmiros (approved):
LGTM

bhautikpip (approved):
Just one suggestion. Should we change the workflow name? like also mention `integration test` just so it's easy to find out integration test workflow ? 

otherwise Looks good:) Thanks for making this change.

anuraaga:
@bhautikpip I'll stick with this name - I think the build for the main branch is the build for the main branch, and that happens to have many different steps, including integration testing.

PR #19, opened 2020-12-04
Don't set version or group in sample apps.

anuraaga:
*Issue #, if available:*

*Description of changes:* We definitely don't want to override version with a constant. And there's no reason to override group either.


bhautikpip (approved):
LGTM :)

PR #26, opened 2020-12-07
Set up sonatype publishing

anuraaga:
*Issue #, if available:*

*Description of changes:* Sets up sonatype publishing target (confirmed can upload locally). Also moves integration test image publishing to after GitHub artifact publishing since they are less important (we wouldn't want to fail other steps if they happen to fail for some reason).

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -188,7 +188,6 @@ allprojects {
       val isSnapshot = version.toString().endsWith("SNAPSHOT")
 
       repositories {
-        // For now, we only publish to GitHub Packages
         maven {
           name = "Sonatype"
           url = uri(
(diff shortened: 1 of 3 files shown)


build.gradle.kts
```diff
@@ -19,6 +19,7 @@ import org.gradle.api.tasks.testing.logging.TestExceptionFormat
 plugins {
   java
   id("nebula.release") version "15.1.0"
+  id("io.codearte.nexus-staging") version "0.22.0"
```
willarmiros, line 22: Why do we need this plugin?

willarmiros (approved):
LGTM

build.gradle.kts
anuraaga: It can allow automating the repo close - for the first time publishing to maven central I'll run the command manually and if it works add it to the workflow.

willarmiros: Neat!

PR #28, opened 2020-12-08
Use SNAPSHOT versioning for dev builds to more easily override JAVA_T…

anuraaga:
…OOL_OPTIONS.

*Issue #, if available:* 

*Description of changes:*

It's hard to override JAVA_TOOL_OPTIONS when not being able to reason about the version number.

Also restores corretto-slim build, 11-alpine-jre was mysteriously unpublished from Docker Hub :O

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -14,6 +14,7 @@
  */
 
 import com.github.jk1.license.render.InventoryMarkdownReportRenderer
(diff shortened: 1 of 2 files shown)


willarmiros (approved):
Huh, pretty surprising that they're taking Corretto images off DockerHub without publishing them to ECR Gallery as a substitute 🤔

PR #31, opened 2020-12-11
Fill up required POM fields.

anuraaga:
*Issue #, if available:*

*Description of changes:* Missing a few fields that maven central will need

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -196,7 +196,7 @@ allprojects {
             scm {
               connection.set("scm:git:git@github.com:aws-observability/aws-otel-java-instrumentation.git")
               developerConnection.set("scm:git:git@github.com:aws-observability/aws-otel-java-instrumentation.git")
-              url.set("git@github.com:aws-observability/aws-otel-java-instrumentation.git")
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -190,6 +192,12 @@ allprojects {
                 roles.add("developer")
               }
             }
+
+            scm {
+              connection.set("scm:git:git@github.com:aws-observability/aws-otel-java-instrumentation.git")
+              developerConnection.set("scm:git:git@github.com:aws-observability/aws-otel-java-instrumentation.git")
+              url.set("git@github.com:aws-observability/aws-otel-java-instrumentation.git")
```
willarmiros, line 199: This isn't consistent with the URL for the Java SDK: https://github.com/aws/aws-xray-sdk-java/blob/master/build.gradle.kts#L270

Should it be a proper URL or is this format ok too?

willarmiros (approved):

build.gradle.kts
anuraaga: Good callout - I copied from instrumentation but agree there doesn't seem to be a reason to copy the exact same value all of three times.

PR #32, opened 2020-12-12
Add a nightly build using upstream snapshots.

anuraaga:
*Issue #, if available:*

*Description of changes:* Eventually we may have the master build of upstream actually trigger ours, but there are some issues related to permission sharing for that. In the meantime, a nightly provides a lot of assurance, though we have to keep the snapshot versions of the agent, and its dependent SDK, up to date manually.


dependencyManagement/build.gradle.kts
```diff
@@ -21,12 +21,14 @@ plugins {
...
 
+val TEST_SNAPSHOTS = rootProject.findProperty("testUpstreamSnapshots") == "true"
+
 val DEPENDENCY_BOMS = listOf(
   "com.fasterxml.jackson:jackson-bom:2.12.0",
   "com.google.protobuf:protobuf-bom:3.14.0",
   "com.linecorp.armeria:armeria-bom:1.3.0",
   "io.grpc:grpc-bom:1.34.0",
-  "io.opentelemetry:opentelemetry-bom:0.12.0",
+  "io.opentelemetry:opentelemetry-bom:${if (!TEST_SNAPSHOTS) "0.12.0" else "0.12.0"}",
```
willarmiros, line 31: Should this point to SNAPSHOT version on the else case?

.github/workflows/nightly-upstream-snapshot-build.yml
```diff
@@ -0,0 +1,155 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-java@v1
+        with:
+          java-version: 14
+      - uses: gradle/wrapper-validation-action@v1
+      - name: Login to GitHub Package Registry
+        run: echo ${{ secrets.CONTAINER_REGISTRY_TOKEN }} | docker login ghcr.io -u anuraaga --password-stdin
```
willarmiros, line 19: We should probably make a team-wide GH account that we can use in the future instead of tying it to a single user.

willarmiros (approved):
Small comments, thanks!

.github/workflows/nightly-upstream-snapshot-build.yml
anuraaga: Yeah going to migrate to ECR.

dependencyManagement/build.gradle.kts
anuraaga: Nah this needs to be the dependency that the upstream agent defines. It's annoying to maintain this, but it's probably ok (I might try a download POM and parse approach later).

PR #35, opened 2021-01-20
Update to OTel 0.14

anuraaga:
*Issue #, if available:*

*Description of changes:*
Had to work around issue that prevents SPI configuration to work until 0.15.0. Fix is in upstream in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2077 but that has breaking changes to configuration options which we can't push out in a patch. This workaround is tame enough.

### DIFF
--- a/awsagentprovider/src/main/java/io/opentelemetry/sdk/trace/RandomIdGenerator.java
+++ b/awsagentprovider/src/main/java/io/opentelemetry/sdk/trace/RandomIdGenerator.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package io.opentelemetry.sdk.trace;
+
+import io.opentelemetry.api.trace.SpanId;
+import io.opentelemetry.sdk.extension.aws.trace.AwsXrayIdGenerator;
+import java.util.concurrent.ThreadLocalRandom;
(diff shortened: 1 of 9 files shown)


awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsTracerProviderFactory.java
```diff
@@ -15,13 +15,12 @@
...
-import io.opentelemetry.api.OpenTelemetry;
-import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.api.trace.TracerProvider;
 import io.opentelemetry.sdk.extension.aws.trace.AwsXrayIdGenerator;
 import io.opentelemetry.sdk.trace.SdkTracerProvider;
-import io.opentelemetry.spi.OpenTelemetryFactory;
+import io.opentelemetry.spi.trace.TracerProviderFactory;
 
-public class AwsOpenTelemetryFactory implements OpenTelemetryFactory {
+public class AwsTracerProviderFactory implements TracerProviderFactory {
```
anuraaga, line 23: Unnecessary changes, these will change again in 0.15.0 to `implements TracerProviderConfigurer`

awsagentprovider/src/main/java/io/opentelemetry/sdk/trace/RandomIdGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+
+  @Override
+  public String generateSpanId() {
+    // Inline implementation since AwsXrayIdGenerator would delegate back to RandomIdGenerator
+    // causing an infinite loop.
+    long id;
+    ThreadLocalRandom random = ThreadLocalRandom.current();
+    do {
+      id = random.nextLong();
+    } while (id == INVALID_ID);
```
willarmiros, line 38: jw, why a `do..while` loop here?

willarmiros (approved):
LGTM

awsagentprovider/src/main/java/io/opentelemetry/sdk/trace/RandomIdGenerator.java
anuraaga: Copied this from upstream - it's just to avoid the random being `0` which is invalid, even though the probability of that is extremely low.

PR #36, opened 2021-01-21
Switch from ghcr to ecr

anuraaga:
*Issue #, if available:*

*Description of changes:*
Was still using ghcr here, but let's finally switch to ecr. The URL doesn't have to look great for these testing images so it's just a default random alias name

### DIFF
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Login to ECR
         uses: docker/login-action@v1
         with:
-          registry: public.ecr.ews
+          registry: public.ecr.aws
       - name: Build snapshot with Gradle
         uses: burrunan/gradle-cache-action@v1
         with:
@@ -68,7 +68,7 @@ jobs:
       - name: Login to ECR
         uses: docker/login-action@v1
         with:
-          registry: public.ecr.ews
+          registry: public.ecr.aws
 
       - name: Setup and Run OTel Collector
         run: |
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -14,14 +14,16 @@ jobs:
         with:
           java-version: 14
       - uses: gradle/wrapper-validation-action@v1
-      - name: Login to GitHub Package Registry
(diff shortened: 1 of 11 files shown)


.github/workflows/release-build.yml
```diff
@@ -55,22 +60,3 @@ jobs:
...
-        with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          aws-region: us-west-2
-
-      - name: Login to Amazon ECR
-        id: login-ecr
-        uses: aws-actions/amazon-ecr-login@v1
-
-      - name: Push image to Amazon ECR
```
willarmiros, line 70: Using the `jib` command does this push instead now right?

willarmiros (approved):
LGTM!

.github/workflows/main-build.yml
```diff
@@ -14,14 +14,16 @@ jobs:
...
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-west-2
       - name: Login to ECR
         uses: docker/login-action@v1
         with:
-          registry: ${{ secrets.AWS_ACCOUNT }}.dkr.ecr.us-west-2.amazonaws.com
-          username: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          password: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          registry: public.ecr.ews
```
lupengamzn, line 26: Shouldn't this be `public.ecr.aws`?

anuraaga: It is, isn't it?

lupengamzn: I mean it's `ews` currently

anuraaga: Doh thought I copy-pasted this from somewhere but apparently not, thanks for seeing it!

.github/workflows/release-build.yml
anuraaga: Yeah

PR #41, opened 2021-02-03
create code scan workflow

wyTrivail:
as per security requested

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,67 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    # The branches below must be a subset of the branches above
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,67 @@
...
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
```
anuraaga, line 17: I don't think we usually run on PR - doesn't the result get posted to the security tab on the repo, not the PR?

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,67 @@
...
+    # The branches below must be a subset of the branches above
+    branches: [ main ]
+  schedule:
```
anuraaga, line 20: Ah also don't think there's a point in a schedule if we check every push anyways, just wastes GitHub's resources.

.github/workflows/codeql-analysis.yml
wyTrivail: sure, this workflow was generated by github, i just removed the conditions on pr since codeql can not block PR in a bad result.

.github/workflows/codeql-analysis.yml
wyTrivail: removed.

anuraaga (approved):
Thanks - can you make the same change to all the repos?

wyTrivail:
> Thanks - can you make the same change to all the repos?

sure, thx

PR #43, opened 2021-02-25
Add publish-plugin

anuraaga:
*Description of changes:*
This is the successor to the gradle-nexus-staging-plugin we previously tried but couldn't due to conflicts with jib. It's modern code and doesn't conflict so we should finally be able to automate the nexus close.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -22,6 +22,7 @@ plugins {
 
   id("com.diffplug.spotless")
   id("com.github.jk1.dependency-license-report")
+  id("io.github.gradle-nexus.publish-plugin")
   id("nebula.release")
 }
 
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -22,15 +22,28 @@ plugins {
...
+      snapshotRepositoryUrl.set(uri("https://aws.oss.sonatype.org/content/repositories/snapshots/"))
+      username.set(System.getenv("PUBLISH_USERNAME"))
+      password.set(System.getenv("PUBLISH_PASSWORD"))
+    }
+  }
+}
+
 val releaseTask = tasks.named("release")
 val postReleaseTask = tasks.named("release")
+val closeAndReleaseTask = tasks.named("closeAndReleaseSonatypeStagingRepository")
```
willarmiros, line 46: Where do we use this?

willarmiros (approved):
Excited to see this work + bring it to other repos

build.gradle.kts
anuraaga: Didn't need it in the end, nice catch

PR #49, opened 2021-03-16
Add instrumentation to inject x-ray formatted IDs into log4j and logb…

### LINKED ISSUE #38: Trace id added to MDC is misleading
Hello,

**Describe the bug**
We use the java agent with one of our spring boot apps. The agent adds a traceId like `60108d573a343aec0c7a35f4ede1e064` to our log output but it should have been something like `1-60108d57-3a343aec0c7a35f4ede1e064`.

**What did you expect to see?**
Maybe the proper traceId copy&paste'ble to Xray? Im not quite sure what is the expected behavior here.

**Additional context**
v.0.6.0 sidecar and v0.12.1-aws.1 agent

Is this expected/work in progress or do I miss something?

Thanks

anuraaga:
…ack contexts.

*Issue #, if available:* Fixes #38

*Description of changes:*
This allows adding `%X{AWS-XRAY-TRACE-ID}` as per https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-java-configuration.html#xray-sdk-java-configuration-logging

/cc @trask @pavolloffay

### DIFF
--- a/instrumentation/logback-1.0/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/logback_1_0/AwsXrayLoggingEventInstrumentation.java
+++ b/instrumentation/logback-1.0/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/logback_1_0/AwsXrayLoggingEventInstrumentation.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.opentelemetry.javaagent.instrumentation.logback_1_0;
+
+import static io.opentelemetry.javaagent.tooling.bytebuddy.matcher.AgentElementMatchers.implementsInterface;
+import static java.util.Collections.singletonMap;
+import static net.bytebuddy.matcher.ElementMatchers.isMethod;
+import static net.bytebuddy.matcher.ElementMatchers.isPublic;
+import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.takesArguments;
+
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanContext;
+import io.opentelemetry.instrumentation.logback.v1_0.internal.UnionMap;
+import io.opentelemetry.javaagent.instrumentation.api.InstrumentationContext;
+import io.opentelemetry.javaagent.tooling.TypeInstrumentation;
+import java.util.Collections;
+import java.util.Map;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.implementation.bytecode.assign.Assigner.Typing;
+import net.bytebuddy.matcher.ElementMatcher;
+
+public class AwsXrayLoggingEventInstrumentation implements TypeInstrumentation {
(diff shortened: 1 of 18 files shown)


instrumentation/logback-1.0/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/logback_1_0/AwsXrayLoggingEventInstrumentation.java
```diff
@@ -0,0 +1,97 @@
...
+    public static void onExit(
+        @Advice.This ILoggingEvent event,
+        @Advice.Return(typing = Typing.DYNAMIC, readOnly = false) Map<String, String> contextData) {
+      if (contextData != null && contextData.containsKey(TRACE_ID_KEY)) {
+        // Assume already instrumented event if traceId is present.
+        return;
+      }
+
+      Span currentSpan = InstrumentationContext.get(ILoggingEvent.class, Span.class).get(event);
+      if (currentSpan == null || !currentSpan.getSpanContext().isValid()) {
```
willarmiros, line 74: We only inject in the case of sampled traces in the X-Ray SDK, do we want to keep the same behavior here?

instrumentation/logback-1.0/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/logback_1_0/AwsXrayLoggingEventInstrumentation.java
```diff
@@ -0,0 +1,97 @@
...
+        return;
+      }
+
+      SpanContext spanContext = currentSpan.getSpanContext();
+      String value =
+          TRACE_ID_KEY
+              + ": 1-"
```
willarmiros, line 81: Originally in the X-Ray SDK, `AWS-XRAY-TRACE-ID` was the default prefix. But customers thought it was unnecessary and verbose, so we made it configurable and disabled by default in the X-Ray agent. I'm not sure how easy it would be to extend the OTel agent configuration in our AWS distro to include such a configuration option, but even if we can't do that I still think we should drop the prefix.

willarmiros (commented):
A couple comments I left on the Logback instr, although they'd apply to the Log4J instr as well.

instrumentation/logback-1.0/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/logback_1_0/AwsXrayLoggingEventInstrumentation.java
anuraaga: So I guess backend features would work without the prefix too (assuming it must if customers are tweaking it)? In that case yeah I'll remove it, the best way to configure the prefix is to add it to the logging config itself.

instrumentation/logback-1.0/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/logback_1_0/AwsXrayLoggingEventInstrumentation.java
anuraaga: I think we may as well to match the behavior of injection of the upstream agent too.

trask:
nice distro extension 👍

instrumentation/logback-1.0/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/logback_1_0/AwsXrayLoggingEventInstrumentation.java
willarmiros: Ok that's fair too. OTel agent should do OTel things

instrumentation/logback-1.0/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/logback_1_0/AwsXrayLoggingEventInstrumentation.java
willarmiros: Yep, ServiceLens just searches for the trace ID and doesn't care what's around it (not even the `@spanId` bit, but we should keep that).

anuraaga: Cool - removed the prefix

willarmiros (approved):
lgtm

PR #50, opened 2021-04-07
Add an AWS composite propagator that can be used in AWS-service compa…

anuraaga:
…tible projects.

*Issue #, if available:*

*Description of changes:*

A composite propagator that allows extracting from several different formats to support compatibility with a variety of client applications, even ones with legacy app instrumentation. It can be configured to either inject with the extracted format, or with the most AWS-compatible format currently, the AWS XRay propagator.

### DIFF
--- a/awspropagator/src/main/java/software/amazon/opentelemetry/awspropagator/AwsCompositePropagator.java
+++ b/awspropagator/src/main/java/software/amazon/opentelemetry/awspropagator/AwsCompositePropagator.java
@@ -1,0 +1,124 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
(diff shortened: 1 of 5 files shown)


awspropagator/src/main/java/software/amazon/opentelemetry/awspropagator/AwsCompositePropagator.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  @Override
+  public <C> void inject(Context context, C carrier, TextMapSetter<C> setter) {
+    if (injectExtractedFormat) {
+      TextMapPropagator extractedPropagator = context.get(EXTRACTED_PROPAGATOR);
+      if (extractedPropagator != null) {
+        extractedPropagator.inject(context, carrier, setter);
+        Baggage baggage = Baggage.fromContextOrNull(context);
+        if (baggage != null && extractedPropagator != AwsXrayPropagator.getInstance()) {
```
willarmiros, line 96: So only the X-Ray propagator supports baggage as part of the trace header?

willarmiros (approved):
lgtm

awspropagator/src/main/java/software/amazon/opentelemetry/awspropagator/AwsCompositePropagator.java
anuraaga: The only other one I know of is jaeger propagator - if we add support for jaeger propagator than this could would need a tweak but for our current list it works well.

PR #72, opened 2021-08-19
Wire up xray sampler.

anuraaga:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smoke-tests/runner/src/test/java/io/awsobservability/instrumentation/smoketests/runner/SpringBootSmokeTest.java
+++ b/smoke-tests/runner/src/test/java/io/awsobservability/instrumentation/smoketests/runner/SpringBootSmokeTest.java
@@ -225,8 +225,8 @@ class SpringBootSmokeTest {
     }
 
     var exported = getExported();
-    // 5 spans per request (1 CLIENT, 2 SERVER, 2 INTERNAL). The default sampler is around 5%, we do
-    // a very rough check that there are no more than 10% sampled.
+    // 5 spans per request (1 CLIENT, 2 SERVER, 2 INTERNAL) if sampled. The default sampler is
+    // around 5%, we do a very rough check that there are no more than 10% sampled.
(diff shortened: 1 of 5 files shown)


smoke-tests/runner/src/test/java/io/awsobservability/instrumentation/smoketests/runner/SpringBootSmokeTest.java
```diff
@@ -186,6 +201,35 @@ void hello() {
...
+    // fetch rules and fallback to the default, which will not sample all requests.
+    WebClient client =
+        WebClient.of("http://localhost:" + applicationXraySampler.getMappedPort(8080));
+    int numRequests = 100;
+    for (int i = 0; i < numRequests; i++) {
+      client.get("/hello").aggregate().join();
+    }
+
+    var exported = getExported();
+    // 5 spans per request (1 CLIENT, 2 SERVER, 2 INTERNAL). The default sampler is around 5%, we do
```
willarmiros, line 228: Is sampling in otel applied to every span individually regardless of Kind or parent? I would expect it to only be applied to the root span like what X-Ray does.

willarmiros (approved):

smoke-tests/runner/src/test/java/io/awsobservability/instrumentation/smoketests/runner/SpringBootSmokeTest.java
anuraaga: Yeah OTel has the ParentBased sampler, which is the default that has that behavior. So 5% is applied to the root which then will either enable or disable the rest of the spans in the trace. I added the word "if sampled" to my comment, does it make it less confusing?

willarmiros: Ah yep that clarifies it

PR #81, opened 2021-09-20
Remove `-test-` component for Sample App Images

NathanielRN:
*Description of changes:*

Both of `springboot` and the `spark` app have containers with a `-test-` component in that last part of their image name.

https://github.com/aws-observability/aws-otel-java-instrumentation/blob/30f9f95629a71397b1ab90643f5b9d8bd133c241/sample-apps/springboot/build.gradle.kts#L18

https://github.com/aws-observability/aws-otel-java-instrumentation/blob/30f9f95629a71397b1ab90643f5b9d8bd133c241/sample-apps/spark/build.gradle.kts#L26

This PR adds it to the `spark-awssdkv1` app so that everything is consistent.

### DIFF
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -83,7 +83,7 @@ jobs:
         env:
           INSTANCE_ID: ${{ github.run_id }}-${{ github.run_number }}
           LISTEN_ADDRESS: 0.0.0.0:8080
-          APP_IMAGE: public.ecr.aws/aws-otel-test/aws-otel-java-test-springboot:${{ github.sha }}
+          APP_IMAGE: public.ecr.aws/aws-otel-test/aws-otel-java-springboot:${{ github.sha }}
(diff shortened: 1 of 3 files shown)


willarmiros (approved):

anuraaga:
I think my intention was to remove the `-test` from all of them at some point, for some of those images names it predates having the public ECR alias with `-test` inside of it. But we need to use Terraform to actually set up the ECR repos before making this change

NathanielRN:
Sounds good! I've made a PR in the terraform repo and updated this PR to remove all the `-test-` components instead.

anuraaga:
@NathanielRN I have merged in the change to migrate to federated login, can you update to HEAD (which doesn't have the main-build-test.yml anymore)?

NathanielRN:
@anuraaga sounds good! I've rebased onto the latest `main` :+1:

anuraaga (approved):

PR #82, opened 2021-09-22
Update to Otel 1.6

anuraaga:
*Issue #, if available:*

*Description of changes:*


awsagentprovider/build.gradle.kts
```diff
@@ -24,7 +24,7 @@ base {
 
 dependencies {
   compileOnly("io.opentelemetry.javaagent:opentelemetry-javaagent-extension-api")
-  compileOnly("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure")
+  compileOnly("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure-spi")
```
willarmiros, line 27: Is this a breaking change from upstream?

willarmiros (approved):
LGTM

awsagentprovider/build.gradle.kts
anuraaga: The SPI interfaces were moved to a separate artifact, which would also be available through the `autoconfigure` package. But we don't need anything but the interfaces so pointed at the more precise one here.

willarmiros: Ah I see it migrated from alpha. Makes sense thanks

PR #83, opened 2021-09-22
Add `SAMPLE_APP_LOG_LEVEL` env var to control sample app logging

NathanielRN:
# Description

For the incoming Soak Tests, we would like to add an environment to control when the Sample Apps produces log outputs. We use branching to control the log messages that indicate an API endpoint was hit, and we remove the `System.out.println` log messages from the `MetricEmitter.java` file.

### DIFF
--- a/sample-apps/spark/src/main/java/com/amazon/sampleapp/App.java
+++ b/sample-apps/spark/src/main/java/com/amazon/sampleapp/App.java
@@ -16,6 +16,8 @@ import software.amazon.awssdk.services.s3.S3Client;
 
 public class App {
   private static final Logger logger = LogManager.getLogger();
+  private static final boolean shouldSampleAppLog =
+      System.getenv().getOrDefault("SAMPLE_APP_LOG_LEVEL", "INFO").equals("INFO");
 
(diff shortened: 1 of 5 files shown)


sample-apps/spark/src/main/java/com/amazon/sampleapp/App.java
```diff
@@ -55,7 +55,9 @@ public static void main(String[] args) {
     get(
         "/outgoing-http-call",
         (req, res) -> {
-          logger.info("Executing outgoing-http-call");
+          if (System.getenv("SAMPLE_APP_LOG_LEVEL") == "INFO") {
```
anuraaga, line 58: Create a `private static final boolean sampleAppLog = "info".equals(System.getenv("SAMPLE_APP_LOG_LEVEL"))` to avoid evaluating this every time.

And think lowercase is better for the value
And we need to use `.equals`

NathanielRN: Spoke offline that `INFO` upper case works well too. Thanks for the help! Changes made!

anuraaga (approved):

PR #89, opened 2021-09-28
Use GitHub OIDC Call to get AWS Creds before STS Call

NathanielRN:
# Description

Follow up to #88, before we make a call to `aws sts get-session-token`, we need to have _some_ AWS credentials. We don't want the credentials to be long-lived, so we try to use the `AWS_WEB_IDENTITY_TOKEN_FILE` env variable resolution method to provide credentials. GitHub will populate this file with temporary credentials (that last 5 minutes) that will give us enough time to call `sts` for additional temporary credentials which last longer (24 hours).

### DIFF
--- a/.github/workflows/soak-testing.yml
+++ b/.github/workflows/soak-testing.yml
@@ -99,6 +99,14 @@ jobs:
 
       - name: Configure AWS Credentials
         run: |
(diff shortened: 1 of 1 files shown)


wangzlei (approved):
LGTM

PR #90, opened 2021-09-28
Use Long Lasting AWS Creds for Soak Tests only

NathanielRN:
# Description

In #88 and #89, we learned that there was _no way_ to get short-lived credentials for the Soak Tests.

If we use GitHub OIDC alone, the tokens expire after 5 minutes and credentials are unrecoverable if not refreshed within that time. Soak Tests run for 5 hours and at a collection interval of 10 minutes so they don't refresh in time.

If we use `aws sts get-session-token` we still need long lasting credentials. As mentioned in [the aws CLI docs](https://docs.aws.amazon.com/cli/latest/reference/sts/get-session-token.html):

> The GetSessionToken operation must be called by using the long-term Amazon Web Services security credentials of the Amazon Web Services account root user or an IAM user.

Trying to use GitHub Sessional Credentials to get additional sessional STS credentials results in [the following error](https://github.com/aws-observability/aws-otel-java-instrumentation/runs/3735504769?check_suite_focus=true):

```
   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  1429    0  1429    0     0  18088      0 --:--:-- --:--:-- --:--:-- 18088

An error occurred (AccessDenied) when calling the GetSessionToken operation: Cannot call GetSessionToken with session credentials
Error: Process completed with exit code 254.
```

So for Soak Tests, we need to use long lasting credentials.

# Alternative Solutions to this PR

It is possible that because this is such a new feature, GitHub might update their token expiration to last longer in the future in which case we can go back to that authentication.

We might consider making a separate role for Soak Tests with more strict permissions, but this still requires long last credentials.

Finally, we might considering **forcing the Soak Tests to refresh every 5 minutes** before the token expires. Then we could use GitHub OIDC, but it would require changing the Soak Test code.

### DIFF
--- a/.github/workflows/soak-testing.yml
+++ b/.github/workflows/soak-testing.yml
@@ -36,7 +36,6 @@ jobs:
     name: Soak Performance Test - (${{ matrix.app-platform }}, ${{ matrix.instrumentation-type }})
     runs-on: ubuntu-latest
     permissions:
-      id-token: write
       contents: write
       issues: write
     strategy:
@@ -98,19 +97,13 @@ jobs:
     # MARK: - Run Performance Tests
 
       - name: Configure AWS Credentials
-        run: |
-          export AWS_ROLE_ARN=${{ secrets.AWS_ASSUME_ROLE_ARN }}
-          export AWS_WEB_IDENTITY_TOKEN_FILE=/tmp/awscreds
-
-          echo "AWS_ROLE_ARN=$AWS_ROLE_ARN" >> $GITHUB_ENV
-          echo "AWS_WEB_IDENTITY_TOKEN_FILE=$AWS_WEB_IDENTITY_TOKEN_FILE" >> $GITHUB_ENV
-
-          curl -H "Authorization: bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" "$ACTIONS_ID_TOKEN_REQUEST_URL&audience=sigstore" | jq -r '.value' > $AWS_WEB_IDENTITY_TOKEN_FILE
-
-          AWS_CREDENTIALS=$(aws sts get-session-token)
-          echo "AWS_ACCESS_KEY_ID=$(echo $AWS_CREDENTIALS | jq '.Credentials.AccessKeyId')" >> $GITHUB_ENV;
-          echo "AWS_SECRET_ACCESS_KEY=$(echo $AWS_CREDENTIALS | jq '.Credentials.SecretAccessKey')" >> $GITHUB_ENV;
-          echo "AWS_SESSION_TOKEN=$(echo $AWS_CREDENTIALS | jq '.Credentials.SessionToken')" >> $GITHUB_ENV;
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          role-to-assume: ${{ secrets.AWS_ASSUME_ROLE_ARN }}
+          role-duration-seconds: 21600 # 6 Hours
+          aws-region: ${{ env.AWS_DEFAULT_REGION }}
       # NOTE: We only login to prevent getting throttled for too many docker
       # pulls. We do not publish anything to ECR.
       - name: Login to ECR


anuraaga:
For the record, I've generally been negative on the soak tests since they seem to provide no real data - I can't imagine an actionable item coming out of them, especially since we may not be able to determine any offending commits if there is a problem. So it's disappointing that they currently block removing the long term IAM key, though admittedly the inability to set the expiration of GITHUB_TOKEN is also limiting there.

Instrumentation is deterministic - 1000 requests per sec in 1 min will always expose any problems that doing so for hours would, since there's no state. This may be motivation to reconsider the approach, replacing these with shorter run tests that run every commit - by being 1:1 with commits at least those would allow some sort of triaging potential.

anuraaga (approved):
In the meantime we may as well make this change since it's what we do in all the other repos. Hoping to remove the IAM user soon though :)

NathanielRN:
Hm maybe it would be worth it to write up a `README` that goes along with these Soak Tests to explain their value 🙂 Although would probably be wise to do that once we figure out a way to put all the Soak Test logic in one place (an action perhaps?) Something like that which had a demo might be useful to show how these tests are useful!

In the meantime let me try to answer your concerns 🙂 

>  I've generally been negative on the soak tests since they seem to provide no real data

This was a major goal for the Soak Tests: To provide real data on host metrics when the Instrumentation SDKs are used. I would say we accomplished this by automatically updating [a graph visualization that shows performance per commit like on Python](https://aws-observability.github.io/aws-otel-python/soak-tests/per-commit-overall-results/index.html):

image

Anonymous users can answer the question "What will the CPU Load Overhead be for my Flask app?" by referencing this graph. In Python we also profile an app _without_ instrumentation and the results quantitatively prove that instrumentation has a negligible overhead even after 5 hours which I thought was an exciting finding. (Although GitHub runners make it hard to rely on this for now but we're working on getting more stable machines for reproducible results 😄)

You can download these results (or get them directly from [the gh-pages branch](https://github.com/aws-observability/aws-otel-python/tree/gh-pages/soak-tests)) by clicking this button:

image

We identified this as important data our users care about similar to [the performance results that are generated by the collector](https://github.com/aws-observability/aws-otel-collector/blob/main/docs/performance_model.md). While those tests only run for 10 minutes, the way these Soak Tests were written lend themselves well to be used as shorter Overhead Tests on every commit because everything is just in a `docker-compose`! We're not there yet but we can be with a lot of the work already done 😄

Only after doing doing the Soak Tests do we have the follow stats for each language:

```
Python:

- CPU Load: 71
- Memory Usage: 2.5 GiB


.NET:

- CPU Load: 70
- Memory Usage: 16 GiB


Go:

- CPU Load: 50
- Memory Usage: 1 GiB


JS:

- CPU Load: 65
- Memory Usage: 1.5 GiB


Java:

- CPU Load: 40
- Memory Usage: 4 GiB
```

It reveals non-obvious things like that the .NET app uses a TON of memory and Go and Java don't use much at all.

> I can't imagine an actionable item coming out of them, especially since we may not be able to determine any offending commits if there is a problem.

You **should** be able to determine the offending commit! The graph I mentioned above is _only_ updated when the Soak Tests are run on a **new** commit. Additionally, _every_ run's results over the 5 hour soak test are posted automatically on the `gh-pages` [branch in folders separated by their commit SHA like on Python](https://github.com/aws-observability/aws-otel-python/tree/gh-pages/soak-tests/snapshots/commits) where you can see the data that caused abnormal CPU/Memory stats.

In addition to providing all the data, the Soak Tests will automatically create an Issue with all this data in a helpful place! For example consider [this issue that was automatically raised on the .NET repo](https://github.com/aws-observability/aws-otel-dotnet/issues/12). It contains information about the exact commit that caused the alert, the images that go along with it, and links to the workflow run + gh-pages branch with the data.

If the Soak Tests averages cause the graph to spike by 175%, the Soak tests will also automatically create an issue saying that **this specific commit** caused a performance hike which would allow us to immediately narrow down and investigate!

In addition to automatic Issue filing per commit for threshold violates DURING and AFTER the Soak Tests, we made some serious modifications to the GitHub action to allow the Soak Tests to be run **manually from any commit**. You'll notice that the Soak Test workflow has a `workflow_dispatch` option which allows you to input a suspicious commit and run the Soak Tests again so that maintainers can verify if this was a one time issue by a commit or otherwise reproduce the error and view the logs/snapshots/CW data to help in debugging the problem 🙂 

> So it's disappointing that they currently block removing the long term IAM key, though admittedly the inability to set the expiration of GITHUB_TOKEN is also limiting there.

I offered one suggestion, such as forcing the Soak Tests to refresh every 5 minutes in code. It seems like we're just very ahead on implementing this feature, maybe we could reach out to the GitHub team about this use case? I think they would be happy that we're using actions for this automated purpose 🙂 

> Instrumentation is deterministic - 1000 requests per sec in 1 min will always expose any problems that doing so for hours would, since there's no state. This may be motivation to reconsider the approach, replacing these with shorter run tests that run every commit - by being 1:1 with commits at least those would allow some sort of triaging potential.

I think this is a very natural progression as I mentioned above 🙂 I regretted that I didn't investigate into [the vegeta HTTP load testing tool](https://github.com/tsenart/vegeta) more when we talked about it casually, because I see that they have [a rate parameter](https://github.com/tsenart/vegeta#-rate) which would allow us to easily set the TPS during the Soak Tests which is a limitation of the `hey` tool we use right now (although `hey` has a prettier report format 😝 ).

Howeverrr, I think that's a small 1 line of code change to implement! All the effort we did to make the Soak Tests run in `docker-compose` files and environment variables means we really just need to add a `THROUGHPUT_PER_SECOND` variable and modify the `DEFAULT_TEST_DURATION_MINUTES` value to 10 minutes instead of 300 minutes.

https://github.com/aws-observability/aws-otel-java-instrumentation/blob/f0c2595b9e4c095f9b017d1c72ac73511900a09e/.github/workflows/soak-testing.yml#L23

https://github.com/aws-observability/aws-otel-java-instrumentation/blob/f0c2595b9e4c095f9b017d1c72ac73511900a09e/.github/workflows/soak-testing.yml#L28-L32

You can actually already configured this parameter on the `worklow_dispatch`:

https://github.com/aws-observability/aws-otel-java-instrumentation/blob/f0c2595b9e4c095f9b017d1c72ac73511900a09e/.github/workflows/soak-testing.yml#L8-L16

Regarding the length not being the most useful, I've noticed that it takes a few minutes to get things to show up in the Soak Tests. I'll list a few examples.

Not to pick on .NET, but it takes almost 90 minutes for it to reach it's peak CPU load:

CPU Load

Or while it's Memory Usage is very high at the beginning, it eventually goes down and an average over the 5 hours better represents its usage:

Memory Usage

Notice the Python "none" app is pretty consistent on Memory usage:

None Py Memory Usage

but the instrumented Manual app has a lot more spikes in memory:

Memory Manual py Spikes

> In the meantime we may as well make this change since it's what we do in all the other repos. Hoping to remove the IAM user soon though :)

I think these are valuable to keep, but definitely agree that as part of the iterative process goes they can be improved 😄 I hope we can find an answer to the IAM user as well! We can bring this up during standup and discuss the options I mentioned and potentially others!

anuraaga:
I agree that the performance tests and much of the code makes sense - it's mostly the concept of long running soak tests for instrumentation that need to be reconsidered I think.

Some things to keep in mind

- If tests are run periodically, rather than on every commit, then it's not possible to `create an issue saying that this specific commit caused a performance hike`. It could be one of several commits and will cause confusion to anyone watching the repository and getting issue notifications.

- There are an infinite number of combinations of instrumentation in a user app - in the end of the day, it's just not possible to really make claims about the overhead we will add to a user app. We can provide some numbers for our limited use cases, and showing performance trends over commits is something, but in the end as a user it's generally not possible to take these numbers / graphs in to make any practical decisions when working back from the customer.

Instrumentation authors traditionally avoid publishing performance numbers for these above reasons, as well as general number stability even when using cloud VMs, and often they actually cause more confusion than benefit. I personally think it's still good to have the numbers - but it is definitely a lot more nuanced than any number is a useful one and I would generally not want to publicize them too much.

An interesting innovation here could be augmenting the test framework used by the languages to instead of unit testing, run the same code in a load test manner. That would at least give good coverage of different frameworks.

- dotnet seems like an outlier, though I'd need to confirm that higher load over a short period of time. It looks like it's probably behavior of the dotnet runtime though, not instrumentation. We have to again work back from the experience of the maintainer this time - given these graphs, what's a maintainer supposed to do? I don't think action items are clear. I suspect that shorter running tests will be similar.

So these are things to keep in mind while improving these, I encourage looking at the tests analytically and with respect to actual user flows. I agree that most of the infrastructure in place can probably be used, though replacing sample apps with actual instrumentation tests is a cool idea that  might diverge somewhat. In the meantime, I believe that we will be in a better state with the relatively simple modification of having these run per commit for a short time instead of being long running tests.

PR #110, opened 2021-11-02
Use aws-credentials action to configure creds bc it uses OIDC by default

NathanielRN:
*Issue #, if available:*

N/A

*Description of changes:*

GitHub just announced [their OIDC feature as GA](https://github.blog/changelog/2021-10-27-github-actions-secure-cloud-deployments-with-openid-connect/), but what’s even cooler is that they seem to have worked with the “configure-aws-credentials GH Action” team to integrate it seamlessly into the action :]

From the [aws-actions/configure-aws-credentials updated README](https://github.com/aws-actions/configure-aws-credentials#assuming-a-role) from 15 days ago:

> We recommend using GitHub’s OIDC provider to get short-lived credentials needed for your actions. Specifying role-to-assume without providing an aws-access-key-id  or a web-identity-token-file will signal to the action that you wish to use the OIDC provider. The default session duration is 1 hour when using the OIDC provider to directly assume an IAM Role. The default session duration is 6 hours when using an IAM User to assume an IAM Role (by providing an aws-access-key-id, aws-secret-access-key, and a role-to-assume) . If you would like to adjust this you can pass a duration to role-duration-seconds, but the duration cannot exceed the maximum that was defined when the IAM Role was created. The default session name is GitHubActions, and you can modify it by specifying the desired name in role-session-name.

This should mean we can use the action to do OIDC authentication, and that the Soak Tests can also use it because the 1 hour default expiry will be enough for it to refresh those credentials 🙂 Previously this was not possible because the OIDC token had a default expiry of 5 minutes.

### DIFF
--- a/.github/workflows/soak-testing.yml
+++ b/.github/workflows/soak-testing.yml
@@ -99,8 +99,6 @@ jobs:
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v1
         with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           role-to-assume: ${{ secrets.AWS_ASSUME_ROLE_ARN }}
           role-duration-seconds: 21600 # 6 Hours
           aws-region: ${{ env.AWS_DEFAULT_REGION }}
--- a/.github/collector/docker-compose.yml
+++ b/.github/collector/docker-compose.yml
@@ -4,8 +4,10 @@ services:
     image: amazon/aws-otel-collector:latest
     command: --config /config/collector-config.yml
     environment:
+      - AWS_ACCESS_KEY_ID
+      - AWS_SECRET_ACCESS_KEY
+      - AWS_SESSION_TOKEN
       - AWS_ROLE_ARN
-      - AWS_WEB_IDENTITY_TOKEN_FILE
     volumes:
       - .:/config
       - /tmp/awscreds:/tmp/awscreds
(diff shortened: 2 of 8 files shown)


.github/workflows/soak-testing.yml
```diff
@@ -99,8 +99,6 @@ jobs:
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v1
         with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
willarmiros, line 103: So these were for a user due to the previous OIDC limitations, but now we can remove the user and these long term creds right?

willarmiros (approved):
Wow, it's like they read our wishlist!

Just to confirm my understanding - we will be using the new default of the configure-aws-credentials which uses the GitHub OIDC auth under the hood. This will mean 1 hour assumed role duration for all actions (which is more than sufficient) except the soak tests, which we configure to 6 hours. Correct?

Also will want to wait for @anuraaga's approval here.

.github/workflows/soak-testing.yml
NathanielRN: Use it was for a user, correct!

Well originally we could not use OIDC because [the token only lasted for 5 minutes](https://github.com/aws-observability/aws-otel-java-instrumentation/pull/90). OIDC is able to refresh the sessional credentials on every AWS call, but the Soak Tests might wait up to 10 minutes to make another AWS call.

The hope is that now, with a 1 hour expiry, that is plenty of time for the Soak Tests to refresh credentials every time it polls the alarms (every 10 minutes) to see if a spike occurred. If the Soak Tests ever poll longer than that then we would have a bigger problem...

However, I have not tested this on my soak-testing-demo fork and was counting on just merging it and fixing it if it breaks 😅... I can move this file to a separate PR if we want to test it on just the Integration Tests first 🙂

anuraaga (approved):
Cool, let's try this!

anuraaga:
Perhaps the new code hasn't been released yet?

https://github.com/aws-observability/aws-otel-java-instrumentation/runs/4099357345?check_suite_focus=true

Let me revert this for now since I'll probably be releasing a patch today and need main build up

willarmiros:
> Perhaps the new code hasn't been released yet?
> 
> https://github.com/aws-observability/aws-otel-java-instrumentation/runs/4099357345?check_suite_focus=true
> 
> Let me revert this for now since I'll probably be releasing a patch today and need main build up

Lol yep https://github.com/aws-actions/configure-aws-credentials/issues/302

PR #130, opened 2021-12-16
GH Action OIDC AWS Creds for GH workflows

NathanielRN:
*Description of changes:*

Trying #110 again now that the action has released the changes as minor release to `v1`! In hindsight we could've tagged a commit on the https://github.com/aws-actions/configure-aws-credentials repository to have the changes right away, but this way we avoid the PR to move them all to be `@v1` again later.

I tested these changes on my fork, where I saw the `Configure AWS credentials` step succeed, but the test overall failed due to permissions needed to push to a public ECR repo. Here is a screenshot of the `main-build.yml` working, with the other workflows being very similar:

image

Soak Testing failed at first, but eventually worked when I added the `id-token: write` permissions to the workflow file:

image

Soak Test eventually ran again and succeeded after 5 hours: https://github.com/NathanielRN/aws-otel-java-instrumentation/actions/runs/1592857236

### DIFF
--- a/.github/workflows/docker-build-corretto-slim.yml
+++ b/.github/workflows/docker-build-corretto-slim.yml
@@ -8,7 +8,6 @@ on:
 
 env:
   AWS_DEFAULT_REGION: us-east-1
-  AWS_WEB_IDENTITY_TOKEN_FILE: /tmp/awscreds
 
 permissions:
   id-token: write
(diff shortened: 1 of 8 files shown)


anuraaga:
I believe we need to make a small change to the IAM config (change from sigstore to sts.amazonaws.com). I can probably get to that early next week let's wait a bit if that's OK

NathanielRN:
@anuraaga Awesome thanks so much! Let me know if I can help, and next week is fine there's no rush 🙂

.github/workflows/docker-build-corretto-slim.yml
```diff
@@ -23,11 +22,10 @@ jobs:
 
       - run: sleep 5 # there's still a race condition for now
```
anuraaga, line 23: Would be cool if you can followup checking if this is still needed, probably not.

anuraaga (approved):

.github/workflows/docker-build-corretto-slim.yml
NathanielRN: I pushed this to my fork which has OIDC set up and it logged in without a problem, except that I think the workflow failed because I don't have authority to push the tag as a last step: https://github.com/NathanielRN/aws-otel-java-instrumentation/actions/runs/1612693300

So I made PR #134 for this change.

Just curious, did we need to sleep because we can't run multiple `sts:AssumeRoleWithWebIdentity` too close to each other?

PR #134, opened 2021-12-22
Remove Sleep from Docker Corretto Workflow

NathanielRN:
*Issue #, if available:* N/A

*Description of changes:*

Following up with https://github.com/aws-observability/aws-otel-java-instrumentation/pull/130#discussion_r772858253, since we use GitHub OIDC to log in now, we should not need this sleep in our GH workflow. We used to do this because the AWS Creds GH action was in an alpha stage and failed without it. Now it is released and ready to be used.

### DIFF
--- a/.github/workflows/docker-build-corretto-slim.yml
+++ b/.github/workflows/docker-build-corretto-slim.yml
@@ -19,8 +19,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: gradle/wrapper-validation-action@v1
-
-      - run: sleep 5 # there's still a race condition for now
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v1
         with:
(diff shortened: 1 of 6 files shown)


anuraaga:
Ah sorry I didn't mean just the one workflow, we have a ton of `sleep`s

https://github.com/aws-observability/aws-otel-java-instrumentation/search?q=sleep

The sleep was in the original snippet that got popular during the feature's beta (alpha?) period, and indeed when I tried without it it failed. It was likely because the runner didn't block correctly waiting for the token to be prepared, and we can be confident that such a race condition would have been fixed by now as it's a released feature.

NathanielRN:
> we can be confident that such a race condition would have been fixed by now as it's a released feature.

That makes total sense! Sorry for the delay, I've updated the PR to remove the sleep from all the workflows 🙂

anuraaga (approved):

PR #146, opened 2022-01-25
Use Opentelemetry SDK For Spark Sample App

sethAmazon:
*Description of changes:*
As of the newer versions of the api. It currently uses a no op provider. This will allow us to use the sdk provider so we can send metrics via grpc. 

### DIFF
--- a/sample-apps/spark/build.gradle.kts
+++ b/sample-apps/spark/build.gradle.kts
@@ -13,6 +13,8 @@ dependencies {
   implementation("org.apache.logging.log4j:log4j-core")
   implementation("software.amazon.awssdk:s3")
   implementation("software.amazon.awssdk:sts")
+  implementation("io.opentelemetry:opentelemetry-sdk-metrics:1.10.0-alpha")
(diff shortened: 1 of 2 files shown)


sample-apps/spark/build.gradle.kts
```diff
@@ -13,6 +13,8 @@ dependencies {
   implementation("org.apache.logging.log4j:log4j-core")
   implementation("software.amazon.awssdk:s3")
   implementation("software.amazon.awssdk:sts")
+  implementation("io.opentelemetry:opentelemetry-sdk-metrics:1.10.0-alpha")
```
anuraaga, line 16: The sample app includes the javaagent so shouldn't need the SDK nor is that really an allowed configuration. Is something not working right now with the javaagent?

sethAmazon: The issue is we are not publishing anything to the collector since the api build a no op provider.

sethAmazon:
Closing as there is a flag to turn on otlp export

PR #162, opened 2022-04-01
Update spring dependency

ruthvik17:
*Issue #, if available:*

Spring RCE issue: https://tanzu.vmware.com/security/cve-2022-22965

*Description of changes:*
Updating the Spring boot versions to [2.6.6](https://spring.io/blog/2022/03/31/spring-boot-2-6-6-available-now).
Update the Spring dependency in sample apps using [a workaround](https://github.com/snicoll-scratches/spring-boot-cve-2022-22965/blob/main/spring-boot-2.4-gradle/build.gradle#L16-L18).


sample-apps/springboot/build.gradle.kts
```diff
@@ -4,6 +4,10 @@ plugins {
   id("com.google.cloud.tools.jib")
 }
 
+
+//This property override Spring Framework to the version that contains the fix
+ext['spring-framework.version'] = '5.3.18'
```
Aneurysm9, line 9: We don't need this if we have updated to spring-boot `v2.6.6`, do we?  I'd not want to see this become stale unnecessarily.

vasireddy99: I am not 100% sure, but for building the `sample-app`
image here. it is not clear for me either which version of `spring-framework` it is built off. overriding to the fixed version should be okay I guess?. 

Also from [info](https://spring.io/blog/2022/03/31/spring-framework-rce-early-announcement) and [example](https://github.com/snicoll-scratches/spring-boot-cve-2022-22965/blob/main/spring-boot-2.4-gradle/build.gradle#L16-L18)
 
image

vasireddy99: I am not 100% sure, but for building the `sample-app`
image here. it is not clear for me either which version of `spring-framework` it is built off. overriding to the fixed version should be okay I guess?. 

Also from [info](https://spring.io/blog/2022/03/31/spring-framework-rce-early-announcement) and [example](https://github.com/snicoll-scratches/spring-boot-cve-2022-22965/blob/main/spring-boot-2.4-gradle/build.gradle#L16-L18)
 
image

ruthvik17: That makes sense. I thought since I couldn't explicitly mention the version, I'd add the code to override the Spring version

Aneurysm9 (approved):

PR #163, opened 2022-04-07
Updating fasterxml jackson library to fix vulnerability

jinwoov:
*Issue #, if available:* 
N/A
*Description of changes:*
This is to update the Jackson library vulnerability. By updating the version, the databind nested dependency will be updated to version `2.13.2.1`, which will have fixed version.

Reference Link: https://mvnrepository.com/artifact/com.fasterxml.jackson/jackson-bom/2.13.2
image

### DIFF
--- a/dependencyManagement/build.gradle.kts
+++ b/dependencyManagement/build.gradle.kts
@@ -31,7 +31,7 @@ val otelSnapshotVersion = "1.13.0"
 
 val DEPENDENCY_BOMS = listOf(
   "com.amazonaws:aws-java-sdk-bom:1.12.185",
-  "com.fasterxml.jackson:jackson-bom:2.13.2",
+  "com.fasterxml.jackson:jackson-bom:2.13.2.1",
   "com.google.guava:guava-bom:31.1-jre",
   "com.google.protobuf:protobuf-bom:3.19.4",
   "com.linecorp.armeria:armeria-bom:1.14.1",
--- a/dependencyManagement/build.gradle.kts
+++ b/dependencyManagement/build.gradle.kts
@@ -31,7 +31,7 @@ val otelSnapshotVersion = "1.13.0"
 
 val DEPENDENCY_BOMS = listOf(
   "com.amazonaws:aws-java-sdk-bom:1.12.185",
-  "com.fasterxml.jackson:jackson-bom:2.13.2.1",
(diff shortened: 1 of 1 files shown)


dependencyManagement/build.gradle.kts
```diff
@@ -31,7 +31,7 @@ val otelSnapshotVersion = "1.13.0"
 
 val DEPENDENCY_BOMS = listOf(
   "com.amazonaws:aws-java-sdk-bom:1.12.185",
-  "com.fasterxml.jackson:jackson-bom:2.13.2",
+  "com.fasterxml.jackson:jackson-bom:2.13.2.1",
```
vasireddy99, line 34: Thank you so much for the PR and detailed [description](https://mvnrepository.com/artifact/com.fasterxml.jackson/jackson-bom/2.13.2)
, it looks like the updated version is `2.13.2.2`. can we possibly use that version.

jinwoov: Sure! updated the jackson bom to [2.13.2.20220328](https://mvnrepository.com/artifact/com.fasterxml.jackson/jackson-bom/2.13.2.20220328)

jinwoov: image

vasireddy99: ⭐ :star:

vasireddy99 (approved):

dependencyManagement/build.gradle.kts
jinwoov: @vasireddy99 Quick question on this, so I can loop back to my team. When is this version going to be available 😅

anuraaga: Hi @jinwoov - the version specified in this repo is actually only used in tests, the actual update will happen after the next upstream javaagent release and then release of the ADOT version. This should happen in 1-2 weeks. Note that if you aren't using the javaagent with AWS Lambda, you aren't using the agent's jackson-databind and would be unaffected

jinwoov: Hey @anuraaga, thanks for the update. I'll keep an eye out for the release. The issue with lower version of jackson-databind is more to do with Prisma picking up the vulnerability in the cluster. Hopefully new release will solve this problem. 😄

Issue #170, opened 2022-05-02
make software.amazon.opentelemetry:aws-opentelemetry-agent default option available as jar instead of pom in mavencentral


gurudatta-carbon:
**Describe the bug**
for the **aws opentelemetry javaagent**, artifacts available in maven central repository
`software.amazon.opentelemetry:aws-opentelemetry-agent`, it's by default available as pom artifact, instead of jar artifact.
https://mvnrepository.com/artifact/software.amazon.opentelemetry/aws-opentelemetry-agent/1.13.1

if you observe gradle representation, by default pom extension is there
```
// https://mvnrepository.com/artifact/software.amazon.opentelemetry/aws-opentelemetry-agent
implementation group: 'software.amazon.opentelemetry', name: 'aws-opentelemetry-agent', version: '1.13.1', ext: 'pom'
```
<img width="345" alt="image" src="https://user-images.githubusercontent.com/81990237/166260693-90b571ef-b780-4223-98e4-9c330f4a7e12.png">

When compared with **io.opentelemetry.javagent**, it's available as jar in mavencentral.
https://mvnrepository.com/artifact/io.opentelemetry.javaagent/opentelemetry-javaagent/1.9.0
if you observe gradle representation, by default jar extension is there
```
// https://mvnrepository.com/artifact/io.opentelemetry.javaagent/opentelemetry-javaagent
runtimeOnly group: 'io.opentelemetry.javaagent', name: 'opentelemetry-javaagent', version: '1.9.0'

```
<img width="463" alt="image" src="https://user-images.githubusercontent.com/81990237/166260736-e5ff7037-00b3-4cd2-b852-729802fe755f.png">

**Steps to reproduce**
- if we don't give extension as jar, it won't download the jar for aws otel java agent

below is working one
`   implementation group: 'software.amazon.opentelemetry', name: 'aws-opentelemetry-agent', version: '1.9.0', ext: 'jar'`
below get's pom by default
`   implementation group: 'software.amazon.opentelemetry', name: 'aws-opentelemetry-agent', version: '1.9.0'`

**What did you expect to see?**

- would like to have jar as the default option for aws opentelemetry agent
**What did you see instead?**
- seeing pom as the default option

**Additional context**

daschneider:
We're running into this issue as well.  If possible please fix.

rapphil:
The fix for this issue will be available in v1.19.0.

rapphil:
[...]

PR #228, opened 2022-10-08
Publish to maven using pom with jar packaging

rapphil:
**IMPORTANT**: We should review the implications of this change before merging. DONT MERGE YET.

Solves #170

**Description of changes:**

Changes the artifact that is resolved from a dependency created to `software.amazon.opentelemetry:aws-opentelemetry-agent` to the jar.

### DIFF
--- a/otelagent/build.gradle.kts
+++ b/otelagent/build.gradle.kts
@@ -100,6 +100,12 @@ tasks {
       attributes.put("Implementation-Version", implementationVersion)
     }
   }
+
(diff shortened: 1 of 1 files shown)


humivo (approved):

bryan-aguilar (approved):

vasireddy99 (approved):

rapphil:
After discussion we deliberated that this change is backwards compatible for existing cases:

* users resolving with the dependency with the `@` for the extension.
* new users that will opt to use the dependency without specifying the extension.

PR #240, opened 2022-10-17
Update dependencies and Otel to v1.19.0

rapphil:
**Description of changes:** Update dependencies to release v1.19.0

Obs.:
* org.slf4j 2.0.3 is backwards compatible with the last version published in the 1.x series. The only difference is that 2.0.3 requires Java 8.
* Updating jacksom-bom to 2.14.0-rc2 because it is only used in tests.

### DIFF
--- a/dependencyManagement/build.gradle.kts
+++ b/dependencyManagement/build.gradle.kts
@@ -27,24 +27,24 @@ data class DependencySet(val group: String, val version: String, val modules: Li
 val TEST_SNAPSHOTS = rootProject.findProperty("testUpstreamSnapshots") == "true"
 
 // This is the only version that affects the released artifact.
-val otelVersion = "1.18.0"
-val otelSnapshotVersion = "1.19.0"
+val otelVersion = "1.19.0"
+val otelSnapshotVersion = "1.20.0"
 
 // All versions below are only used in testing and do not affect the released artifact.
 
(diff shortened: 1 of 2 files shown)


dependencyManagement/build.gradle.kts
```diff
@@ -27,24 +27,24 @@ data class DependencySet(val group: String, val version: String, val modules: Li
...
+val otelVersion = "1.19.0"
+val otelSnapshotVersion = "1.20.0"
 
 // All versions below are only used in testing and do not affect the released artifact.
 
 val DEPENDENCY_BOMS = listOf(
-  "com.amazonaws:aws-java-sdk-bom:1.12.321",
-  "com.fasterxml.jackson:jackson-bom:2.13.4",
+  "com.amazonaws:aws-java-sdk-bom:1.12.322",
+  "com.fasterxml.jackson:jackson-bom:2.14.0-rc2",
```
Aneurysm9, line 37: Should this be using an `rc`?

rapphil: It was in the summary: Updating jacksom-bom to 2.14.0-rc2 because it is only used in tests.

Aneurysm9: Reading is hard!

Aneurysm9 (approved):

PR #242, opened 2022-10-18
Cleanup for release v1.19.0

rapphil:
*Description of changes:* 
* Update jacksom bom to the latest stable version.
* Revert update in org.slf4j dependency. This was breaking integration tests because the test application stopped working as intended.
* Updated all tests and workflows to point to the registry `aws-otel-test`
* Create `integrationTests` task that is able to run locally exclusively so that the pr-build gh action can be idempotent. (previously it was depending on artifacts generated during the main-build gh action). This required adding some specific logic in the `jib` plugin because local docker daemon does not handle multiplatform images.

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -20,8 +20,14 @@ jobs:
         with:
           java-version: 17
       - uses: gradle/wrapper-validation-action@v1
-      - name: Build with Gradle
+      - name: Build with Gradle with Integration tests
+        if: ${{ matrix.os == 'ubuntu-latest' }}
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: build --stacktrace -PenableCoverage=true -PlocalDocker=true
+          arguments: build integrationTests --stacktrace -PenableCoverage=true -PlocalDocker=true
+      - name: Build with Gradle
+        uses: gradle/gradle-build-action@v2
+        if: ${{ matrix.os != 'ubuntu-latest' }}
+        with:
+          arguments: build --stacktrace -PenableCoverage=true
       - uses: codecov/codecov-action@v1
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -23,5 +23,5 @@ jobs:
       - name: Build with Gradle
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: build --stacktrace -PenableCoverage=true
+          arguments: build --stacktrace -PenableCoverage=true -PlocalDocker=true
       - uses: codecov/codecov-action@v1
--- a/dependencyManagement/build.gradle.kts
+++ b/dependencyManagement/build.gradle.kts
@@ -34,7 +34,7 @@ val otelSnapshotVersion = "1.20.0"
 
 val DEPENDENCY_BOMS = listOf(
   "com.amazonaws:aws-java-sdk-bom:1.12.322",
-  "com.fasterxml.jackson:jackson-bom:2.14.0-rc2",
+  "com.fasterxml.jackson:jackson-bom:2.13.4.20221013",
   "com.google.guava:guava-bom:31.1-jre",
   "com.google.protobuf:protobuf-bom:3.21.7",
   "com.linecorp.armeria:armeria-bom:1.20.1",
@@ -60,7 +60,7 @@ val DEPENDENCY_SETS = listOf(
   ),
   DependencySet(
     "org.slf4j",
-    "2.0.3",
+    "1.7.36",
     listOf(
       "slf4j-api",
       "slf4j-simple"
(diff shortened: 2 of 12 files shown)


Aneurysm9 (approved):

dependencyManagement/build.gradle.kts
```diff
@@ -60,7 +60,7 @@ val DEPENDENCY_SETS = listOf(
   ),
   DependencySet(
     "org.slf4j",
-    "2.0.3",
+    "1.7.36",
```
vasireddy99, line 63: As we are facing any issue using the latest [SLF4J API Module](https://mvnrepository.com/artifact/org.slf4j/slf4j-api). just wondering if it might have to do with this : https://www.slf4j.org/faq.html#changesInVersion200

sample-apps/spark-awssdkv1/build.gradle.kts
```diff
@@ -19,27 +21,20 @@ application {
...
-  }
+  )
 }
 
 tasks {
   named("jib") {
     dependsOn(":otelagent:jib")
   }
+  named("jibDockerBuild") {
+    dependsOn(":otelagent:jibDockerBuild")
```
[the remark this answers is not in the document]
rapphil: No, docker daemon is still not able to support multi arch. This change is just forcing us to use jibDockerBuild in integration tests so that we don't need to rely on the artifacts that are published in the main-build Github CI.

sample-apps/springboot/build.gradle.kts
```diff
@@ -14,27 +16,21 @@ dependencies {
 }
 
 jib {
-  to {
-    image = "public.ecr.aws/aws-otel-test/aws-otel-java-springboot"
+  configureImages(
+    "public.ecr.aws/aws-otel-test/aws-opentelemetry-java-base:alpha",
+    "public.ecr.aws/aws-otel-test/aws-otel-java-springboot",
+    rootProject.property("localDocker")!!.equals("true"),
```
bryan-aguilar, line 22: Excuse my gradle ignorance but is there a reason that these don't match the changes in other gradle build files by explicitly stating the parameter they are targetting? 
```
   localDocker = rootProject.property("localDocker")!!.equals("true"),
    multiPlatform = !rootProject.property("localDocker")!!.equals("true"),
```

.github/workflows/pr-build.yml
```diff
@@ -20,8 +20,14 @@ jobs:
...
           java-version: 17
       - uses: gradle/wrapper-validation-action@v1
+      - name: Build with Gradle with Integration tests
+        if: ${{ matrix.os == 'ubuntu-latest' }}
+        uses: gradle/gradle-build-action@v2
+        with:
+          arguments: build integrationTests --stacktrace -PenableCoverage=true -PlocalDocker=true
       - name: Build with Gradle
         uses: gradle/gradle-build-action@v2
+        if: ${{ matrix.os != 'ubuntu-latest' }}
```
bryan-aguilar, line 30: Why are macos and windows tests executed with different gradle inputs than an ubuntu runner?

bryan-aguilar, line 30: Why are macos and windows tests executed with different gradle inputs than an ubuntu runner?

dependencyManagement/build.gradle.kts
rapphil: It is not clear what is the issue a this time, but the entry that you pointed in the faq is related to changes that were introduced in version 2.0 that are new features. I was trusting this statement:

>SLF4J 2.0.0 incorporates an optional [fluent api](https://www.slf4j.org/manual.html#fluent). Otherwise, there are no client facing API changes in 2.0.x. __For most users, upgrading to version 2.0..x should be a drop-in replacement, as long as the logging provider is updated as well.__

.github/workflows/pr-build.yml
rapphil: because they don't support docker.

We were skipping the tests for macos and windows silently because of this:
https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/smoke-tests/runner/src/test/java/io/awsobservability/instrumentation/smoketests/runner/LogInjectionTest.java#L32
```
@Testcontainers(disabledWithoutDocker = true)
class LogInjectionTest {
```

Se we are not running smoke tests in windows and macos anyways. Just building and running some unit tets.

bryan-aguilar: Noted, there are ways we could run docker on MacOS runners. See [here](https://github.com/actions/runner-images/issues/6216). This could possibly be a follow on item if we feel the need to expand test coverage to other operating systems.

bryan-aguilar (approved):

sample-apps/springboot/build.gradle.kts
rapphil: localDocker and multiPlatform applies to the source image. However there are cases where the we want to use a remote image but no multiPlatform. E.g.: for generating this image: aws-opentelemetry-java-base:alpha

PR #321, opened 2023-01-30
Add new process for patch releases

rapphil:
**Description of changes:** Add process for creating patch releases that is able to patch opentelemetry dependencies.

The idea is to checkout the upstream dependencies, apply carefully created patches and publish these patched dependencies to maven local.

### DIFF
--- a/.github/actions/patch-dependencies/action.yml
+++ b/.github/actions/patch-dependencies/action.yml
@@ -2,7 +2,6 @@ name: "Patch dependencies"
 description: |
   Patches direct dependencies of this project leveraging maven local to publish the results.
 
-  This workflow will be noop if there is no dependency to patch, i.e.: no files in `.github/patchs`
   This workflow supports patching opentelemetry-java and opentelemetry-java-instrumentation repositories by executing
   the `patch.sh` script that will try to patch those repositories and after that will optionally test and then publish
   the artifacts to maven local.
@@ -14,6 +13,9 @@ inputs:
     default: "false"
     required: false
     description: "If the workflow should run tests of the dependencies. Anything different than false will evaluate to true"
+  branch:
+    required: true
+    description: "The branch where this patches are being applied e.g.: release/v1.21.x"
   gpg_private_key:
     description: "The gpg key used to sign the artifacts"
     required: true
@@ -25,19 +27,19 @@ runs:
   steps:
     - name: check patches
       run: |
-        if [[ -f .github/patchs/opentelemetry-java.patch ]]; then
+        if [[ -f .github/patches/${{ inputs.branch }}/opentelemetry-java.patch ]]; then
           echo 'patch_otel_java=true' >> $GITHUB_ENV
         fi
-        if [[ -f .github/patchs/opentelemetry-java-instrumentation.patch ]]; then
+        if [[ -f .github/patches/${{ inputs.branch }}/opentelemetry-java-instrumentation.patch ]]; then
             echo 'patch_otel_java_instrumentation=true' >> $GITHUB_ENV
         fi
-        if [[ -f .github/patchs/opentelemetry-java-contrib.patch ]]; then
+        if [[ -f .github/patches/${{ inputs.branch }}/opentelemetry-java-contrib.patch ]]; then
             echo 'patch_otel_java_contrib=true' >> $GITHUB_ENV
         fi
       shell: bash
 
     - name: Clone and patch repositories
-      run: .github/scripts/patch.sh
+      run: .github/scripts/patch.sh "${{ inputs.branch }}"
       if: ${{ env.patch_otel_java == 'true' ||
               env.patch_otel_java_instrumentation == 'true' ||
               env.patch_otel_java_contrib == 'true' }}
--- a/.github/actions/patch-dependencies/action.yml
+++ b/.github/actions/patch-dependencies/action.yml
@@ -1,0 +1,79 @@
+name: "Patch dependencies"
+description: |
+  Patches direct dependencies of this project leveraging maven local to publish the results.
+
+  This workflow will be noop if there is no dependency to patch, i.e.: no files in `.github/patchs`
+  This workflow supports patching opentelemetry-java and opentelemetry-java-instrumentation repositories by executing
+  the `patch.sh` script that will try to patch those repositories and after that will optionally test and then publish
+  the artifacts to maven local.
+  To add a patch you have to add a file in the `.github/patches/` directory with the name of the repository that must
+  be patched.
+  This action assumes that java was set correctly.
+inputs:
+  run_tests:
+    default: "false"
+    required: false
+    description: "If the workflow should run tests of the dependencies. Anything different than false will evaluate to true"
+  gpg_private_key:
+    description: "The gpg key used to sign the artifacts"
+    required: true
+  gpg_password:
+    description: "The gpg key password"
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - name: check patches
+      run: |
+        if [[ -f .github/patchs/opentelemetry-java.patch ]]; then
+          echo 'patch_otel_java=true' >> $GITHUB_ENV
+        fi
+        if [[ -f .github/patchs/opentelemetry-java-instrumentation.patch ]]; then
+            echo 'patch_otel_java_instrumentation=true' >> $GITHUB_ENV
+        fi
+      shell: bash
+
+    - name: Clone and patch repositories
+      if: ${{ env.patch_otel_java == 'true' && env.patch_otel_java_instrumentation == 'true' }}
+      run: .github/scripts/patch.sh
+      shell: bash
+
+    - name: Build opentelemetry-java with tests
+      uses: gradle/gradle-build-action@v2
+      with:
+        arguments: build publishToMavenLocal
+        build-root-directory: opentelemetry-java
+      env:
+        GPG_PRIVATE_KEY: ${{ inputs.gpg_private_key }}
+        GPG_PASSWORD: ${{ inputs.gpg_password }}
+      if: ${{ env.patch_otel_java == 'true' && inputs.run_tests != 'false' }}
+
+    - name: Build opentelemetry-java
+      uses: gradle/gradle-build-action@v2
+      with:
+        arguments: publishToMavenLocal
+        build-root-directory: opentelemetry-java
+      env:
+        GPG_PRIVATE_KEY: ${{ inputs.gpg_private_key }}
+        GPG_PASSWORD: ${{ inputs.gpg_password }}
+      if: ${{ env.patch_otel_java == 'true' && inputs.run_tests == 'false' }}
+
+    - name: Build opentelemetry-java-instrumentation with tests
+      uses: gradle/gradle-build-action@v2
+      with:
+        arguments: check -x spotlessCheck publishToMavenLocal
+        build-root-directory: opentelemetry-java-instrumentation
+      env:
+        GPG_PRIVATE_KEY: ${{ inputs.gpg_private_key }}
+        GPG_PASSWORD: ${{ inputs.gpg_password }}
+      if: ${{ env.patch_otel_java_instrumentation == 'true' && inputs.run_tests != 'false' }}
+
+    - name: Build opentelemetry java instrumentation
+      uses: gradle/gradle-build-action@v2
+      with:
+        arguments: publishToMavenLocal
+        build-root-directory: opentelemetry-java-instrumentation
+      env:
+        GPG_PRIVATE_KEY: ${{ inputs.gpg_private_key }}
+        GPG_PASSWORD: ${{ inputs.gpg_password }}
+      if: ${{ env.patch_otel_java_instrumentation == 'true' && inputs.run_tests == 'false' }}
--- a/.github/actions/patch-dependencies/action.yml
+++ b/.github/actions/patch-dependencies/action.yml
@@ -31,49 +31,74 @@ runs:
         if [[ -f .github/patchs/opentelemetry-java-instrumentation.patch ]]; then
             echo 'patch_otel_java_instrumentation=true' >> $GITHUB_ENV
         fi
+        if [[ -f .github/patchs/opentelemetry-java-contrib.patch ]]; then
+            echo 'patch_otel_java_contrib=true' >> $GITHUB_ENV
+        fi
       shell: bash
 
     - name: Clone and patch repositories
-      if: ${{ env.patch_otel_java == 'true' && env.patch_otel_java_instrumentation == 'true' }}
       run: .github/scripts/patch.sh
+      if: ${{ env.patch_otel_java == 'true' OR
+              env.patch_otel_java_instrumentation == 'true' OR
+              env.patch_otel_java_contrib == 'true' }}
       shell: bash
 
     - name: Build opentelemetry-java with tests
       uses: gradle/gradle-build-action@v2
+      if: ${{ env.patch_otel_java == 'true' && inputs.run_tests != 'false' }}
       with:
         arguments: build publishToMavenLocal
         build-root-directory: opentelemetry-java
       env:
         GPG_PRIVATE_KEY: ${{ inputs.gpg_private_key }}
         GPG_PASSWORD: ${{ inputs.gpg_password }}
-      if: ${{ env.patch_otel_java == 'true' && inputs.run_tests != 'false' }}
 
     - name: Build opentelemetry-java
       uses: gradle/gradle-build-action@v2
+      if: ${{ env.patch_otel_java == 'true' && inputs.run_tests == 'false' }}
       with:
         arguments: publishToMavenLocal
         build-root-directory: opentelemetry-java
       env:
         GPG_PRIVATE_KEY: ${{ inputs.gpg_private_key }}
         GPG_PASSWORD: ${{ inputs.gpg_password }}
-      if: ${{ env.patch_otel_java == 'true' && inputs.run_tests == 'false' }}
+
+    - name: Build opentelemetry-java-contrib with tests
+      uses: gradle/gradle-build-action@v2
+      if: ${{ env.patch_otel_java_contrib == 'true' && inputs.run_tests != 'false' }}
+      with:
+        arguments: build publishToMavenLocal
+        build-root-directory: opentelemetry-java-contrib
+      env:
+        GPG_PRIVATE_KEY: ${{ inputs.gpg_private_key }}
+        GPG_PASSWORD: ${{ inputs.gpg_password }}
+
+    - name: Build opentelemetry-java-contrib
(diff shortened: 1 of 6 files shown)


.github/scripts/patch.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash -x
+
+exit 1 # The following values must be change before every new patch release
+OTEL_JAVA_VERSION=1.21.0
```
rapphil, line 4: This has to be updated before every new release. I opted to create this script instead of adding github steps because I have to checkout a specific tag and create a new branch and apply the patch. These set of operations would not go well with the github actions.

.github/actions/patch-dependencies/action.yml
```diff
@@ -0,0 +1,79 @@
...
+  run_tests:
+    default: "false"
+    required: false
+    description: "If the workflow should run tests of the dependencies. Anything different than false will evaluate to true"
+  gpg_private_key:
+    description: "The gpg key used to sign the artifacts"
+    required: true
+  gpg_password:
+    description: "The gpg key password"
+    required: true
```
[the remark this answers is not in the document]
rapphil: https://github.com/aws-observability/aws-otel-java-instrumentation/actions/runs/4041936113/jobs/6949041652#step:5:100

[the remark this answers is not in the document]
rapphil: https://docs.github.com/en/actions/learn-github-actions/contexts#secrets-context
> The secrets context is not available for composite actions due to security reasons. If you want to pass a secret to a composite action, you need to do it explicitly as an input.

.github/actions/patch-dependencies/action.yml
```diff
@@ -0,0 +1,79 @@
...
+        if [[ -f .github/patchs/opentelemetry-java.patch ]]; then
+          echo 'patch_otel_java=true' >> $GITHUB_ENV
+        fi
+        if [[ -f .github/patchs/opentelemetry-java-instrumentation.patch ]]; then
+            echo 'patch_otel_java_instrumentation=true' >> $GITHUB_ENV
+        fi
+      shell: bash
+
+    - name: Clone and patch repositories
+      if: ${{ env.patch_otel_java == 'true' && env.patch_otel_java_instrumentation == 'true' }}
```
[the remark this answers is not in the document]
rapphil: good catch.

.github/actions/patch-dependencies/action.yml
```diff
@@ -0,0 +1,79 @@
...
+
+    - name: Build opentelemetry-java with tests
+      uses: gradle/gradle-build-action@v2
+      with:
+        arguments: build publishToMavenLocal
+        build-root-directory: opentelemetry-java
+      env:
+        GPG_PRIVATE_KEY: ${{ inputs.gpg_private_key }}
+        GPG_PASSWORD: ${{ inputs.gpg_password }}
+      if: ${{ env.patch_otel_java == 'true' && inputs.run_tests != 'false' }}
```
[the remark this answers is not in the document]
rapphil: will change

.github/scripts/patch.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash -x
+
+exit 1 # The following values must be change before every new patch release
```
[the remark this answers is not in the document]
rapphil: This was left on purpose and should be like this on purpose. It is because users need to specify which tag of an upstream dependency they need to patch. this will make sure that they change the file.

.github/workflows/pr-build.yml
```diff
@@ -13,13 +31,23 @@ jobs:
         os:
           - macos-latest
           - ubuntu-latest
-          - windows-latest
+          # Windows is not working for patch workflows, therefore we disable it here
+          # https://github.com/square/wire/issues/2188
+          # https://github.com/open-telemetry/opentelemetry-java/issues/4560
+          # - windows-latest
```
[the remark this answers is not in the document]
rapphil: It is the compilation of opentelemetry-java that is broken in github workflows on windows. This is a deal breaker for the whole patching process.

.github/workflows/main-build.yml
```diff
@@ -3,7 +3,7 @@ on:
   push:
     branches:
       - main
-
+      - "release-*"
```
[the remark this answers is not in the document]
rapphil: > Does the patch process require a branch for every MAJOR.MINOR or a branch for specific patch versions? For example, would the branch be release-v0.21.1 or release-v0.21.x?
The latter. `release-v1.21.x`.

> nit: Should we be using the same format as the collectors release branch naming release/*?
For consistency I think it is fair to change.

.github/actions/patch-dependencies/action.yml
```diff
@@ -0,0 +1,79 @@
...
+    required: true
+  gpg_password:
+    description: "The gpg key password"
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - name: check patches
+      run: |
+        if [[ -f .github/patchs/opentelemetry-java.patch ]]; then
```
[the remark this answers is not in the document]
rapphil: Yes, we can add contrib to the list of dependencies that can be patched.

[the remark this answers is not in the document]
rapphil: The reason why I did not add the contrib one is because there was no good reason for testing it.

.github/scripts/patch.sh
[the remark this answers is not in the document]
rapphil: I added better comments.

.github/workflows/main-build.yml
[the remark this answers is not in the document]
rapphil: btw, I had this to the readme. I updated the readme with the new conventions.

.github/scripts/patch.sh
```diff
@@ -0,0 +1,50 @@
...
+OTEL_JAVA_CONTRIB_PATCH=".github/patchs/opentelemetry-java-contrib.patch"
+if [[ -f "$OTEL_JAVA_CONTRIB_PATCH" ]]; then
+  git clone https://github.com/open-telemetry/opentelemetry-java-contrib.git
+  cd opentelemetry-java-contrib
+  git checkout v${OTEL_JAVA_CONTRIB_VERSION} -b tag-v${OTEL_JAVA_CONTRIB_VERSION}
+  patch -p1 < "../${OTEL_JAVA_CONTRIB_PATCH}"
+  git commit -a -m "Patching to release ${PATCH_VERSION}"
+  cd -
+else
+  echo "Skipping patching opentelemetry-java-instrumentation"
```
bryan-aguilar, line 36:
```suggestion
  echo "Skipping patching opentelemetry-java-contrib"
```

.github/scripts/patch.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash -x
+
+exit 1 # Remove this lock after updating the following values
```
bryan-aguilar, line 3: I don't know why I don't see my previous comment here but I will expand. 

I don't know if this is the correct way to force a user to update the version below. Could this be done by passing in an arg into the script and the validating it? 

If the answer is no and we want to leave this as the process then I think we should add a detailed comment here on the steps required before removing the `exit 1`.

.github/workflows/pr-build.yml
```diff
@@ -13,18 +34,33 @@ jobs:
...
+      - uses: actions/setup-java@v3
         with:
           java-version: 17
+          distribution: temurin
+
       - uses: gradle/wrapper-validation-action@v1
+
+      - name: Publish patched dependencies to maven local
+        uses: ./.github/actions/patch-dependencies
+        if: ${{ contains(github.event.pull_request.base.ref, 'release/v') }}
```
bryan-aguilar, line 53:
```suggestion
        if: ${{ contains(github.event.pull_request.base.ref, 'release/v*') }}
```

.github/scripts/patch.sh
rapphil: > I don't know if this is the correct way to force a user to update the version below. Could this be done by passing in an arg into the script and the validating it?

But from where the argument would come from?

.github/workflows/pr-build.yml
rapphil: No need to add wildcard, but I think instead of contains it is safer to use `startsWith`
https://docs.github.com/en/actions/learn-github-actions/expressions#startswith

.github/scripts/patch.sh
rapphil: I will add the comments.

.github/workflows/pr-build.yml
rapphil: I will fix the other occurrences where I added the wildcard by mistake :(

.github/actions/patch-dependencies/action.yml
```diff
@@ -0,0 +1,104 @@
...
```
Aneurysm9, line 28:
```suggestion
        if [[ -f .github/patches/opentelemetry-java.patch ]]; then
```

.github/actions/patch-dependencies/action.yml
```diff
@@ -0,0 +1,104 @@
...
+        build-root-directory: opentelemetry-java-contrib
+      env:
+        GPG_PRIVATE_KEY: ${{ inputs.gpg_private_key }}
+        GPG_PASSWORD: ${{ inputs.gpg_password }}
+
+    - name: Build opentelemetry-java-instrumentation with tests
+      uses: gradle/gradle-build-action@v2
+      if: ${{ env.patch_otel_java_instrumentation == 'true' && inputs.run_tests != 'false' }}
+      with:
+        arguments: check -x spotlessCheck publishToMavenLocal
```
Aneurysm9, line 90: Is this skipping a check?  Why?  Should there be a comment explaining that inline?

.github/scripts/patch.sh
```diff
@@ -0,0 +1,58 @@
...
+
+# .github/patchs/.versions.sh should define all the versions of the dependencies that we are going to patch
+# This is used so that we can properly clone the upstream repositories.
+# This file should define the following variables:
+# OTEL_JAVA_VERSION. Tag of the opentelemetry-java repository to use. E.g.: JAVA_OTEL_JAVA_VERSION=v1.21.0
+# OTEL_JAVA_INSTRUMENTATION_VERSION. Tag of the opentelemetry-java-instrumentation repository to use, e.g.: OTEL_JAVA_INSTRUMENTATION_VERSION=v1.21.0
+# OTEL_JAVA_CONTRIB_VERSION. Tag of the opentelemetry-java-contrib repository. E.g.: OTEL_JAVA_CONTRIB_VERSION=v1.21.0
+# This script will fail if a variable that is supposed to exist is referenced.
+
+if [[ ! -f .github/patchs/versions ]]; then
```
Aneurysm9, line 11:
```suggestion
if [[ ! -f .github/patches/versions ]]; then
```

.github/scripts/patch.sh
```diff
@@ -0,0 +1,58 @@
...
+  echo "No versions file found. Skipping patching"
+  exit 0
+fi
+
+source .github/patchs/versions
+
+git config --global user.email "github-actions@github.com"
+git config --global user.name "github-actions"
```
Aneurysm9, line 19: Maybe set this to something that makes clear it is this workflow as the actor?

.github/workflows/pr-build.yml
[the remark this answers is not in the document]
Aneurysm9: Does it need to be built on windows to be used?  Can it be built in a separate step that runs on Ubuntu and then used as an artifact in a step running on Windows?  Does building on Windows actually tell us anything given that release artifacts appear to only be built on Ubuntu?  Can we rely on integration tests for Windows?

.github/scripts/patch.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+# Enable debug mode, fail on any command that fail in this script and fail on unset variables
+set -x -e -u
+
+# .github/patchs/.versions.sh should define all the versions of the dependencies that we are going to patch
```
bryan-aguilar, line 5:
```suggestion
# .github/patches/versions should define all the versions of the dependencies that we are going to patch
```

RELEASING.md
```diff
@@ -36,49 +36,31 @@ to view a summary of all commits since last release as a reference.
...
-
-```
-git push origin v1.2.x
-```
+* Provide patch files for each repository that will need to be patched. These files should be located in `.github/patchs` and should be named
+using the convention `<repository name>.patch`. The following repositories are supported: opentelemetry-java, opentelemetry-java-instrumentation and opentelemetry-java-contrib. Provide one patch file per repository. The adot patch version of each upstream dependency should be `<version>+adot<number>` where `version` is the version of the upstream dependency and `number` is the number of this patch that should be incremented from 1 per patch version.
 
-Create a PR to have code review and merge this into upstream's release branch. As this was not
-applied automatically, we need to do code review to make sure the manual cherrypick is correct.
+* Create a `versions` file in the directory `.github/patchs`. This file should contain shell variables with the versions of the tags of the repositories which will receive patchs.
```
bryan-aguilar, line 57:
```suggestion
* Create a `versions` file in the directory `.github/patches`. This file should contain shell variables with the versions of the tags of the repositories which will receive patches.
```

RELEASING.md
```diff
@@ -36,49 +36,31 @@ to view a summary of all commits since last release as a reference.
 All patch releases should include only bug-fixes, and must avoid
 adding/modifying the public APIs. 
 
-Open the patch release build workflow in your browser [here](https://github.com/aws-observability/aws-otel-java-instrumentation/actions?query=workflow%3A%22Patch+Release+Build%22).
+Steps:
+1. Create a branch from the release that you want to patch. It should follow the convention `release/v<major>.<minor>.x`. E.g.: if you want to patch release 1.21.0, the name of the branch should be `release/v1.21.x`.
+1. Mark the branch as protected.
```
bryan-aguilar, line 41: nit: can we use branch protection settings with a wildcard to automatically protect these branches? @Aneurysm9

.github/scripts/patch.sh
```diff
@@ -0,0 +1,60 @@
...
+# OTEL_JAVA_INSTRUMENTATION_VERSION. Tag of the opentelemetry-java-instrumentation repository to use, e.g.: OTEL_JAVA_INSTRUMENTATION_VERSION=v1.21.0
+# OTEL_JAVA_CONTRIB_VERSION. Tag of the opentelemetry-java-contrib repository. E.g.: OTEL_JAVA_CONTRIB_VERSION=v1.21.0
+# This script will fail if a variable that is supposed to exist is referenced.
+
+if [[ ! -f .github/patchs/versions ]]; then
+  echo "No versions file found. Skipping patching"
+  exit 0
+fi
+
+source .github/patchs/versions
```
bryan-aguilar, line 18:
```suggestion
source .github/patches/versions
```

.github/actions/patch-dependencies/action.yml
rapphil: this was copied from upstream. I can add details.

.github/workflows/pr-build.yml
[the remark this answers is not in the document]
rapphil: >Does it need to be built on windows to be used? Can it be built in a separate step that runs on Ubuntu and then used as an artifact in a step running on Windows? 

No, we could build in another job and upload the artifact to github.

>  Does building on Windows actually tell us anything given that release artifacts appear to only be built on Ubuntu?

It tells us:
* That the code compiles.
* That these two UTs pass:
  * https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsTracerConfigurerTest.java
  * https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/awspropagator/src/test/java/software/amazon/opentelemetry/awspropagator/AwsCompositePropagatorTest.java

> Can we rely on integration tests for Windows?

We don't have integration tests running on windows. We never had.

.github/actions/patch-dependencies/action.yml
```diff
@@ -0,0 +1,104 @@
+name: "Patch dependencies"
+description: |
+  Patches direct dependencies of this project leveraging maven local to publish the results.
+
+  This workflow will be noop if there is no dependency to patch, i.e.: no files in `.github/patchs`
```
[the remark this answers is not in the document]
rapphil: This could work if we look for patches based on the branch name.

.github/actions/patch-dependencies/action.yml
rapphil: and it makes sense to leave it off for us because spotlessCheck is a linter that checks for code style issues.

.github/actions/patch-dependencies/action.yml
rapphil: I will fix all occurrences.

.github/workflows/pr-build.yml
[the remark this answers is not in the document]
rapphil: We could add integration tests for windows, but that would require creating windows specific Images since they cannot be based on a Linux base image.

.github/actions/patch-dependencies/action.yml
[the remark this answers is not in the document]
rapphil: Implemented your suggestion. thanks!

.github/scripts/patch.sh
rapphil: done

RELEASING.md
Aneurysm9: Done.  Let me know if there are any issues.  Team members should have the ability to push new branches matching the pattern `release/v*` and the branch will be automatically protected.

.github/actions/patch-dependencies/action.yml
```diff
@@ -25,19 +27,19 @@ runs:
...
         fi
-        if [[ -f .github/patchs/opentelemetry-java-instrumentation.patch ]]; then
+        if [[ -f .github/patches/${{ inputs.branch }}/opentelemetry-java-instrumentation.patch ]]; then
             echo 'patch_otel_java_instrumentation=true' >> $GITHUB_ENV
         fi
-        if [[ -f .github/patchs/opentelemetry-java-contrib.patch ]]; then
+        if [[ -f .github/patches/${{ inputs.branch }}/opentelemetry-java-contrib.patch ]]; then
             echo 'patch_otel_java_contrib=true' >> $GITHUB_ENV
         fi
       shell: bash
```
bryan-aguilar, line 39: nit (Wont block PR): I think this logic could be moved into the `patch.sh` script now that we use a `version` file to determine if patching is necessary. If a patch file exists it could set the output to the github env.

bryan-aguilar (approved):

RELEASING.md
Aneurysm9, line 41:
```suggestion
```

Aneurysm9 (approved):

PR #353, opened 2023-03-15
feat: Add Operator compatible image

rapphil:
*Description of changes:*

* Add copy-utility which is a statically linked rust executable that is used to copy the ADOT Java Agent jar into the Operator volume.
* Add docker image compatible with the images used by the OpenTelemetry operator.

To test locally:

```
./gradlew otelagent:build
docker build -t adot-java-operator:latest --build-arg ADOT_JAVA_VERSION=$(./gradlew printVersion -q) .
docker volume create operator-volume
docker run --mount source=operator-volume,dst=/otel-auto-instrumentation adot-java-operator:latest cp /javaagent.jar /otel-auto-instrumentation/javaagent.jar
docker run -ti --mount source=operator-volume,dst=/otel-auto-instrumentation public.ecr.aws/amazonlinux/amazonlinux:latest
Unable to find image 'public.ecr.aws/amazonlinux/amazonlinux:latest' locally
latest: Pulling from amazonlinux/amazonlinux
Digest: sha256:5d5fbc48e357994e2678d338b2dd1fae74d5f4d5ae2c8c0eae60a48a4f95ce38
Status: Downloaded newer image for public.ecr.aws/amazonlinux/amazonlinux:latest
bash-4.2# ls /otel-auto-instrumentation/
javaagent.jar
```

This is equivalent to what is done here https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/instrumentation/javaagent.go#L71


tools/cp-utility/src/main.rs
```diff
@@ -0,0 +1,338 @@
...
+        process::exit(1);
+    });
+
+    do_copy(operation).unwrap_or_else(|err| {
+        eprintln!("Error copying files: {err}");
+        process::exit(1);
+    });
+}
+
+#[cfg(test)]
```
rapphil, line 131: yes, in rust it is a practice to put unit tests in the same file as in the code: https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-module-and-cfgtest

.github/dependabot.yml
```diff
@@ -9,3 +9,13 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+  - package-ecosystem: "cargo"
+    directory: /tools/cp-utility
+    schedule:
+      interval: "weekly"
+      day: "sunday"
+  - package-ecosystem: "docker"
```
bryan-aguilar, line 17: AFAIK, the `docker` package ecosystem is just for updating image tags in k8s manifests. [docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem).

rapphil: Yes, official docs don't say anything. I was trusting this article: https://dev.to/oracle2025/how-to-keep-a-dockerfile-updated-with-dependabot-1mdn

rapphil: The source code of the dependabot docker package-ecosystem does look for Docker images https://github.com/dependabot/dependabot-core/blob/main/docker/lib/dependabot/docker/file_fetcher.rb

.github/workflows/pr-build.yml
```diff
@@ -67,6 +67,20 @@ jobs:
...
+        run: |
+          echo ADOT_JAVA_VERSION="$(./gradlew printVersion -q )" >> $GITHUB_ENV
+
+      - name: Build image
+        uses: docker/build-push-action@v4
+        if: ${{ matrix.os == 'ubuntu-latest' }}
+        with:
+          push: false
+          build-args: "ADOT_JAVA_VERSION=${{ env.ADOT_JAVA_VERSION }}"
+          context: .
```
PaurushGarg, line 82: Just wondering if we should add the tag here as well?

.github/workflows/pr-build.yml
```diff
@@ -67,6 +67,20 @@ jobs:
         with:
           arguments: build integrationTests --stacktrace -PenableCoverage=true -PlocalDocker=true
 
+      - name: Get current version
```
PaurushGarg, line 70: Are we still planning to pull this pr-build workflow for cp-utility out in a separate workflow file?

.github/workflows/pr-build.yml
rapphil: we are not pushing, so no need to providing tags.

.github/workflows/pr-build.yml
rapphil: I'm debating myself if there is much benefit in separating. I think if we change the tool, we will also want to build the image, which will imply in having to build the jar, therefore we would be duplicating much of what is done in this workflow. 

Do you think it is worth separating? what are the pros? The only thing I can think of is that we would not need to run integration tests.

tools/cp-utility/src/main.rs
```diff
@@ -0,0 +1,338 @@
...
+    let args: Vec<String> = env::args().collect();
+
+    let operation = parse_args(args).unwrap_or_else(|err| {
+        eprintln!("Error parsing arguments: {err}");
+        process::exit(1);
+    });
+
+    do_copy(operation).unwrap_or_else(|err| {
+        eprintln!("Error copying files: {err}");
+        process::exit(1);
```
Aneurysm9, line 127:
```suggestion
        process::exit(2);
```

Distinct error codes for distinct errors.

tools/cp-utility/src/main.rs
```diff
@@ -0,0 +1,338 @@
...
+        io::Write,
+        os::unix,
+        path::{Path, PathBuf},
+    };
+
+    use crate::{do_copy, parse_args, CopyOperation, CopyType};
+    use uuid;
+
+    #[test]
+    fn test_parser() {
```
Aneurysm9, line 144: This only seems to test one argument parsing scenario.  Non-archive copy should be tested as well.

tools/cp-utility/src/main.rs
```diff
@@ -0,0 +1,338 @@
...
+
+/// Execute the recursive type of copy operation
+fn copy_recursive(source: &Path, dest: &Path) -> io::Result<()> {
+    // This will cover the case in which the destination exists
+    let sanitized_dest: PathBuf = if dest.exists() {
+        dest.to_path_buf()
+            .join(source.file_name().ok_or(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "Invalid source
+        file",
```
Aneurysm9, line 84:
```suggestion
                "Invalid source file",
```

rapphil: I'm using rustfmt. I will try to play around with it to check if we can avoid this.

tools/cp-utility/src/main.rs
```diff
@@ -0,0 +1,338 @@
...
+fn do_copy(operation: CopyOperation) -> io::Result<()> {
+    match operation.copy_type {
+        CopyType::Archive => copy_recursive(&operation.source, &operation.destination)?,
+        CopyType::SingleFile => fs::copy(&operation.source, &operation.destination).map(|_| ())?,
+    };
+    Ok(())
+}
+
+/// Execute the recursive type of copy operation
+fn copy_recursive(source: &Path, dest: &Path) -> io::Result<()> {
```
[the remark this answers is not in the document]
rapphil: `cp -a` implies in `cp -r`. Your suggestion makes sense.

.github/workflows/pr-build.yml
PaurushGarg: ok. thanks. I agree this would duplicate the building jar process. Should we add these steps to main-build.yml as well?

.github/workflows/pr-build.yml
```diff
@@ -67,6 +67,20 @@ jobs:
...
+        if: ${{ matrix.os == 'ubuntu-latest' }}
```
PaurushGarg, line 71: if `matrix.is != ubuntu-latest`, then line 84 has `build with gradle` step for that. in that case, would be building the `java agent image` as well?

.github/workflows/pr-build.yml
rapphil: I will add to main build.

.github/workflows/pr-build.yml
rapphil: no. because we will only publish this image for linux. We do not intend to publish this image for other operating systems (I'm not sure if that is even possible).

Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+# Stage 1: Build the cp-utility binary
+FROM rust:1.68 as builder
+
+# possible values: x86_64 | aarch64
+ARG ARCH=x86_64
```
bryan-aguilar, line 19: Where does this get overriden in the build process? Should we be building on both x86/amd64?

tools/cp-utility/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "cp-utility"
+version = "0.1.0"
+edition = "2021"
```
bryan-aguilar, line 4: Is there a reason for 2021 here?

tools/cp-utility/README.md
```diff
@@ -0,0 +1,57 @@
+# Introduction
+
+This is a copy utility that is supposed to be used in a container image used by the
+OpenTelemetry Operator. This copy utility will copy the ADOT Java agent jar from
+the init container to the final destination volume.
```
bryan-aguilar, line 5:
```suggestion
This copy utility is intended to be used as a base image for OpenTelemetry Operator 
autoinstrumentation images. The copy utility will allow the ADOT Java agent jar to be 
copied from the init container to the final destination volume.
```

Dockerfile
bryan-aguilar: Does buildx automatically pass this in for different archs?

tools/cp-utility/src/main.rs
```diff
@@ -0,0 +1,353 @@
...
+    source: PathBuf,
+    /// The destination path
+    destination: PathBuf,
+    /// The type of copy being performed
+    copy_type: CopyType,
+}
+
+/// Parse command line arguments and transform into `CopyOperation`
+fn parse_args(args: Vec<&str>) -> io::Result<CopyOperation> {
+    if !((args.len() == 4 || args.len() == 5 && args[2].eq("-a")) && args[1].eq("cp")) {
```
bryan-aguilar, line 45: Should we expand this error checking? Verify that `args[1] == "cp"` for both arg lengths?

tools/cp-utility/Cargo.toml
rapphil: https://doc.rust-lang.org/edition-guide/editions/index.html
2021 is the latest edition.

tools/cp-utility/src/main.rs
rapphil: oh. it is already checking.
this is the condition that is being checked in negative `(args.len() == 4 || args.len() == 5 && args[2].eq("-a")) && args[1].eq("cp")`

note the `&& args[2].eq("-a")` is considering both the cases of 4 and 5 arguments.

Dockerfile
rapphil: it will build, but we need to pass the `ARCH` argument explicitly. I will modify the workflow.

tools/cp-utility/src/main.rs
bryan-aguilar: Ahh good points. I missed the nested parenthesis.

bryan-aguilar: Can we add a verification test case?

rapphil: good call. I added more tests.

rapphil: https://github.com/aws-observability/aws-otel-java-instrumentation/pull/353/files#diff-255e1973e4371a91e310ea9c1b94c1507bab752b6aa20b39322542e6d205616aR178-R179

bryan-aguilar (approved):

PR #360, opened 2023-03-24
Add workflow to run dependency check based on OWASP

rapphil:
*Description of changes:*
Add dependency check based on OWASP using this gradle plugin: https://plugins.gradle.org/plugin/org.owasp.dependencycheck

### DIFF
--- a/.github/workflows/owasp.yml
+++ b/.github/workflows/owasp.yml
@@ -24,6 +24,12 @@ jobs:
         with:
           name: adot-dependencies
           arguments: ":otelagent:dependencyCheckAnalyze"
+      - name: Upload report
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v3
+        with:
+          name: adot-dependencies
+          path: otelagent/build/reports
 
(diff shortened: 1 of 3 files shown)


bryan-aguilar (approved):

.github/workflows/owasp.yml
```diff
@@ -0,0 +1,63 @@
...
+          fetch-depth: 0
+      - uses: actions/setup-java@v1
+        with:
+          java-version: 17
+      - name: Build snapshot with Gradle
+        uses: gradle/gradle-build-action@v2
+        with:
+          arguments: ":javaagent:dependencyCheckAnalyze"
+      - name: Upload report
+        if: always()
```
bryan-aguilar, line 43: Does this require enclosing in `${{ }}`

rapphil: https://docs.github.com/en/actions/learn-github-actions/expressions

Docs say that it is optional, but for consistency I will add it. good call. thanks!

PR #381, opened 2023-04-14
Fix: Make sure name of artifact is aligned in release workflow

rapphil:
*Description of changes:* This will make sure that the name of the jar is exactly the same during the build for the the docker image and for maven central. 

If we did not make this change the docker step will fail, because it will not be able to find the artifact.


PaurushGarg (commented):
I think this fix would also be required here: https://github.com/aws-observability/aws-otel-java-instrumentation/blob/47e6394d5face8c2caecad6bb1a14b323bb0500f/.github/workflows/patch-release-build.yml#L94

PaurushGarg (approved):

PR #415, opened 2023-05-18
fix release v1.26.x

rapphil:
*Description of changes:* Properly set the region

This was done in the release branch v.1.24.x but was not brought back into main.

### DIFF
--- a/.github/workflows/patch-release-build.yml
+++ b/.github/workflows/patch-release-build.yml
@@ -97,7 +97,7 @@ jobs:
         uses: aws-actions/configure-aws-credentials@v1
         with:
           role-to-assume: ${{ secrets.AWS_ASSUME_ROLE_ARN_RELEASE }}
-          aws-region: us-west-2
+          aws-region: ${{ env.AWS_DEFAULT_REGION }}
 
       - name: Log in to AWS ECR
         uses: docker/login-action@v1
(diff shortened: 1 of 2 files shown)


vasireddy99 (approved):

Kausik-A (approved):

.github/workflows/patch-release-build.yml
```diff
@@ -97,7 +97,7 @@ jobs:
         uses: aws-actions/configure-aws-credentials@v1
         with:
           role-to-assume: ${{ secrets.AWS_ASSUME_ROLE_ARN_RELEASE }}
-          aws-region: us-west-2
+          aws-region: ${{ env.AWS_DEFAULT_REGION }}
```
vasireddy99, line 100: oh, are we running tests in us-east-1 now?

rapphil: this image is not for tests. it is the image containing the ADOT Java Agent used by the operator. It has to use us-east-1 because we are publishing to a public ecr repository.

PaurushGarg (approved):

.github/workflows/patch-release-build.yml
vasireddy99: Thanks for clarifying

PR #489, opened 2023-07-31
[release/v1.28.x] Cache maven local

bryan-aguilar:
Remake of #488 but push to main repo instead of fork.

### DIFF
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -60,13 +60,14 @@ jobs:
         with:
           java-version: 17
           distribution: temurin
-        # we need to cache the local patch outputs across builds
-        # we also only want to restore them if they are an exact match
+          
+        # cache local patch outputs
       - name: Cache local Maven repository
         uses: actions/cache@v3
         with:
-          path: ~/.m2/repository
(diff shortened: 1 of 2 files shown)


.github/workflows/main-build.yml
```diff
@@ -61,6 +61,14 @@ jobs:
           java-version: 17
           distribution: temurin
 
+      - name: Cache local Maven repository
```
rapphil, line 64: this needs to go after we publish everything to maven local no? in line 72

bryan-aguilar: You insert the action at the point in which you want to restore the cache. GitHub actions will automatically cache the path after job success.

.github/workflows/main-build.yml
```diff
@@ -61,6 +61,14 @@ jobs:
...
+        uses: actions/cache@v3
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
```
rapphil, line 68: we don't use `pom.xml`. shouldn't this be `.kts`?

bryan-aguilar: Took another go at the key hash. Opted for a much more targeted hash that included the patch file hash.

rapphil (approved):

PR #601, opened 2023-11-15
Fix default value for when the otel.smp.enabled property is not set

srprash:
*Description of changes:*
The functionality should be explicitly opt-in.

### DIFF
--- a/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsAppSignalsCustomizerProvider.java
+++ b/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsAppSignalsCustomizerProvider.java
@@ -63,7 +63,7 @@ public class AwsAppSignalsCustomizerProvider implements AutoConfigurationCustomi
   }
 
   private boolean isSmpEnabled(ConfigProperties configProps) {
-    return configProps.getBoolean("otel.smp.enabled", true);
+    return configProps.getBoolean("otel.smp.enabled", false);
   }
 
   private Sampler customizeSampler(Sampler sampler, ConfigProperties configProps) {


bryan-aguilar (approved):

mxiamxia (approved):

itstheceo:
Could this change of default value result in a substantial increase of APM data?

srprash:
>Could this change of default value result in a substantial increase of APM data?

Quite the opposite. Setting the `otel.smp.enabled` to `true` will tell the ADOT agent to generate some extra APM data for  [ApplicationSignals](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Application-Monitoring-Sections.html).

itstheceo:
@srprash Thanks for the prompt reply. I have observed a massive increase in APM data for my service over the last 3-4 weeks, though setting `otel.smp.enabled` to `true` appears to be unrelated as you say. There hasn't been much change in this service, mostly bumping dependencies. I am still trying to track down the cause as it's only recently been brought to my attention. From my investigation, fields such as `http.server.duration` are no longer being bucketed into a histogram, which is resulting in a huge increase of documents. I will keep digging.

itstheceo:
Ended up reverting the agent to 1.29.0, and the issue is no longer happening.

srprash:
@itstheceo What was the version of the ADOT agent where you noticed the issue?

itstheceo:
@srprash I suspect something changed in 1.31.x. The issue happened in two services I am responsible for, both had similar symptoms. They do use distributed tracing with each other, whether that's relevant or not. 

It felt like rather than the agent sending the delta, it repeatedly sent everything which was accumulating over time. As you can see following, the subtle gradient trending up. It did eventually level out, assuming when a buffer reached capacity.
Screenshot 2023-12-13 at 11 44 28 AM

And after reverting `aws-opentelemetry-agent` to 1.29.0, traffic returned to normal.
Screenshot 2023-12-13 at 11 44 42 AM

bryan-aguilar:
Thank you @itstheceo,

Was this issue present with both `1.31.0` and `1.31.1`?

itstheceo:
@bryan-aguilar not sure yet. I'll try to set aside some time for testing this week and get back to you.

bryan-aguilar:
That would be helpful, ty. But also, #648

PR #624, opened 2023-12-01
Add retry logic to sample app endpoint connection

harrryr:
*Issue #, if available:*
The E2E canary occasionally fails due to being unable to establish connection with the sample app. 
*Description of changes:*
If connecting to the endpoint fails, then restart the sample app again and reattempt establishing connection; up to 3 times

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #625, opened 2023-12-01
Add retry logic to sample app endpoint connection

harrryr:
*Issue #, if available:*
The E2E canary occasionally fails due to being unable to establish connection with the sample app.
*Description of changes:*
If connecting to the endpoint fails, then restart the sample app again and reattempt establishing connection; up to 3 times

### DIFF
--- a/.github/workflows/appsignals-e2e-eks-test.yml
+++ b/.github/workflows/appsignals-e2e-eks-test.yml
@@ -116,48 +116,50 @@ jobs:
 
       # Application pods need to be restarted for the
       # app signals instrumentation to take effect
-      - name: Restart the app pods
-        run: kubectl delete pods --all -n ${{ env.SAMPLE_APP_NAMESPACE }}
+      - name: Restart the app pods and wait for the endpoints to come online
+        id: restart-pod
+        uses: nick-fields/retry@v2
+        with:
+          max_attempts: 3
+          retry_on: error
+          timeout_minutes: 6
+          shell: bash
+          # Restart the pod, then attempt to connect to the endpoint. It will try 30 times with 10 seconds delay. If
+          # it fails to establish connection, restart the pod and try again up to 3 tries.
+          command: |
+            kubectl delete pods --all -n ${{ env.SAMPLE_APP_NAMESPACE }}
+            kubectl wait --for=condition=Ready pod --all -n ${{ env.SAMPLE_APP_NAMESPACE }}
+            cd testing/terraform/eks
+            sample_app_endpoint=http://$(terraform output sample_app_endpoint)
+            cd ../../../              
+            curl --retry 30 --retry-delay 10 -s -o /dev/null $(echo "$sample_app_endpoint" | tr -d '"')
+            if [ $? != "0" ];then
+              echo "Failed to establish connection with endpoint"
+              exit 1
+            fi
 
-      - name: Wait for sample app pods to come up
-        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/appsignals-e2e-eks-test.yml
```diff
@@ -116,49 +116,51 @@ jobs:
...
       # app signals instrumentation to take effect
-      - name: Restart the app pods
-        run: kubectl delete pods --all -n ${{ env.SAMPLE_APP_NAMESPACE }}
+      - name: Restart the app pods and wait for the endpoints to come online
+        id: restart-pod
+        uses: nick-fields/retry@v2
+        with:
+          max_attempts: 3
+          retry_on: error
+          timeout_minutes: 6
```
srprash, line 125: What is this timeout for? Is it the time limit for the all the commands to run successfully before retrying?

.github/workflows/appsignals-e2e-eks-test.yml
```diff
@@ -116,49 +116,51 @@ jobs:
...
+          shell: bash
+          # Restart the pod, then attempt to connect to the endpoint. It will try 30 times with 10 seconds delay. If
+          # it fails to establish connection, restart the pod and try again up to 3 tries.
+          command: |
```
srprash, line 129: I'm a little confused as to what is the root cause of the transient failure this PR is aiming to fix. Is it that the application pods are occasionally fail to run or is it the ALB endpoint that fails to come up. I feel its the latter.
In that case, does simply restarting the application pods solves the ALB endpoint issue? Have we reproduced/tested this fix anywhere?

.github/workflows/appsignals-e2e-eks-test.yml
harrryr: Yup it's the time limit for all the commands to run. On a successful run it takes less than 2 minutes, but on failed runs it seems like it takes ~6 minutes.

.github/workflows/appsignals-e2e-eks-test.yml
harrryr: Yes, the failure happens when the endpoints fails to come up. We think it's happening due to occasional network problem/timeouts, so in such situations restarting the pod should hopefully allow the endpoints to come up properly. It's hard to reproduce this error, so I'm not 100% sure if it will work, but tested the code to check if the entire workflow succeeds properly and tested to see if the retry logic runs properly in the event that the commands fail.

harrryr:
Switched out of the `nick-fields retry` action after receiving feedback from this comment: https://github.com/aws-observability/aws-otel-java-instrumentation/pull/624#discussion_r1412593230

.github/workflows/appsignals-e2e-eks-test.yml
```diff
@@ -115,50 +115,53 @@ jobs:
...
+            kubectl delete pods --all -n ${{ env.SAMPLE_APP_NAMESPACE }}
+            kubectl wait --for=condition=Ready pod --all -n ${{ env.SAMPLE_APP_NAMESPACE }}
+            sample_app_endpoint=http://$(terraform output sample_app_endpoint)
+            if ! $(curl --retry 30 --retry-delay 10 -s -o /dev/null $(echo "$sample_app_endpoint" | tr -d '"'));then
+              echo "Failed to establish connection with endpoint"
+              retry_counter=$(($retry_counter+1))
+            else
+              success=1
+            fi
+          done
```
majanjua-amzn, line 139: In the worst case, how long will this take? Can we justify taking that long for a single test?

PR #628, opened 2023-12-04
fix-e2e-eks-terraform-destroy-error

harrryr:
*Issue #, if available:*
Although it's not causing functional problems, the `terraform destroy` step on the eks canary returns a failure due to having an invalid aws_region variable. Ex: https://github.com/aws-observability/aws-otel-java-instrumentation/actions/runs/7093413738

*Description of changes:*
Removed aws_region on the terraform provider description

### DIFF
--- a/testing/terraform/eks/main.tf
+++ b/testing/terraform/eks/main.tf
@@ -16,9 +16,7 @@ terraform {
   }
 }
 
-provider "aws" {
-  region = var.aws_region
-}
+provider "aws" {}
 
 # get eks cluster
 data "aws_eks_cluster" "testing_cluster" {


testing/terraform/eks/main.tf
```diff
@@ -16,9 +16,7 @@ terraform {
   }
 }
 
-provider "aws" {
-  region = var.aws_region
-}
+provider "aws" {}
```
majanjua-amzn, line 19: Put this back and instead add `-var="aws_region=${{ env.AWS_DEFAULT_REGION }}" \` in the workflow here: https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/workflows/appsignals-e2e-eks-test.yml#L262

This should work and still maintain the expected behaviour that if someone doesn't explicitly pass in the region it fails.

majanjua-amzn (changes requested):

majanjua-amzn (approved):

PaurushGarg (approved):

PR #634, opened 2023-12-07
Add Retry Mechanism to E2E EKS Terraform Deployment

harrryr:
*Issue #, if available:*
The EKS Canary occasionally fails due to transitivity issues. One of the recurring errors are `Max attempts reached` in the  [step](https://github.com/aws-observability/aws-otel-java-instrumentation/actions/runs/7095916355/job/19313588479) : `Wait for Endpoint to Come Online`. This occurs due to the endpoint sometimes taking longer than expected to become ready. 

*Description of changes:*
- Change the endpoint wait time from 5 minutes to 10 minutes. 
- Add a retry mechanism such that if either the terraform deployment or the endpoint connection fails, then destroy the terraform deployment and try again. This will ensure that the endpoint is deleted and recreated. 

### DIFF
--- a/.github/workflows/appsignals-e2e-eks-test.yml
+++ b/.github/workflows/appsignals-e2e-eks-test.yml
@@ -85,6 +85,7 @@ jobs:
 
       # Enable App Signals on the test cluster
       - name: Pull and unzip enablement script from S3
+        working-directory: testing/terraform/eks
         run: aws s3 cp ${{ env.ENABLEMENT_SCRIPT_S3_BUCKET }} . && unzip -j onboarding.zip
 
       - name: Change ADOT image if main-build
@@ -93,6 +94,7 @@ jobs:
 
       - name: Remove log group deletion command
         if: always()
+        working-directory: testing/terraform/eks
         run: |
           delete_log_group="aws logs delete-log-group --log-group-name '${{ env.LOG_GROUP_NAME }}' --region \$REGION"
           sed -i "s#$delete_log_group##g" clean-app-signals.sh
@@ -137,7 +139,7 @@ jobs:
             # after installing App Signals. Attempts to connect will be made for up to 10 minutes
             if [ $success -eq 0 ]; then
               echo "Installing app signals to the sample app"
-              ../../../enable-app-signals.sh \
+              ./enable-app-signals.sh \
               ${{ inputs.test-cluster-name }} \
               ${{ env.AWS_DEFAULT_REGION }} \
               ${{ env.SAMPLE_APP_NAMESPACE }}
@@ -166,7 +168,7 @@ jobs:
             # resources created from terraform and try again.
             if [ $success -eq 1 ]; then
               echo "Cleaning up App Signal"
-              ../../../clean-app-signals.sh \
+              ./clean-app-signals.sh \
               ${{ inputs.test-cluster-name }} \
               ${{ env.AWS_DEFAULT_REGION }} \
               ${{ env.SAMPLE_APP_NAMESPACE }}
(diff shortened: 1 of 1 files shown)


.github/workflows/appsignals-e2e-eks-test.yml
```diff
@@ -107,21 +91,102 @@ jobs:
...
+          
+            if [ $success -eq 1 ]; then
+              echo "Terraform deployment was unsuccessful. Will attempt to retry deployment."
+            fi
+
+            # If the success is still 0, then the terraform deployment succeeded and now try to connect to the endpoint 
+            # after installing App Signals. Attempts to connect will be made for up to 10 minutes
+            if [ $success -eq 0 ]; then
+              echo "Installing app signals to the sample app"
+              ../../../enable-app-signals.sh \
```
majanjua-amzn, line 140: nit: Can we just pull the enablement script into the folder it would be useful in?

.github/workflows/appsignals-e2e-eks-test.yml
```diff
@@ -139,30 +204,22 @@ jobs:
...
+          curl -S -s -o /dev/null http://${{ env.APP_ENDPOINT }}/outgoing-http-call/
+          curl -S -s -o /dev/null http://${{ env.APP_ENDPOINT }}/aws-sdk-call/
+          curl -S -s -o /dev/null http://${{ env.APP_ENDPOINT }}/remote-service?ip=${{ env.REMOTE_SERVICE_POD_IP }}/
+          curl -S -s -o /dev/null http://${{ env.APP_ENDPOINT }}/client-call/
 
       # Validation for app signals telemetry data
       - name: Call endpoint and validate generated EMF logs
         id: log-validation
-        if: steps.endpoint-check.outcome == 'success' && !cancelled()
+        if: steps.deploy-sample-app.outcome == 'success' && !cancelled()
```
majanjua-amzn, line 222: Do we need to add this to the other validation steps?

majanjua-amzn (approved):
Some small questions but overall should be fine

.github/workflows/appsignals-e2e-eks-test.yml
srprash: +1

srprash:
Is there a sample run where we can see this change in action?

.github/workflows/appsignals-e2e-eks-test.yml
```diff
@@ -107,21 +91,102 @@ jobs:
...
+          retry_counter=0
+          max_retry=2
+          success=1 
+          while [ $success -eq 1 ]; do
+            if [ $retry_counter -ge $max_retry ]; then
+              echo "Max retry reached, failed to deploy terraform and connect to the endpoint. Exiting code"
+              exit 1
+            fi
+            echo "Attempt $retry_counter"
+            success=0
```
srprash, line 120: Success should be made 0 *after* the `terraform apply` command has completed. That's is the assumption in the following code.

harrryr: Success of 1 indicates that the setting up App Signals on the sample app failed, while 0 indicates that everything ran successfully. 

The logic is:
1. Set Success to 1 (Set initial value to 1 so that the while loop runs)
2. While Success is 1 (Indicates that terraform deployment/endpoint connection failed and will try again):
     2a:  Set Success to 0 (Set the value to 0 and if there were any failures change it to 1)
     2b: Run Terraform apply (If the deployment failed, then success will change to 1)
     2c: If Success is still 0, then install App Signals and check endpoint connection
     2d: If endpoint connection failed, change success to 1
     2e: If Success is 1 at this point, then either the deployment or connection failed and run the while loop again.       If it is still 0, then the code ran successfully and exit the while loop

If the success is made 0 after the terraform apply, then it will override whether terraform deployment succeeded or not. If after the terraform deployment the success is 1, we want to skip the endpoint connection step and redeploy the terraform again.

majanjua-amzn: I think it's a little confusing because 1 usually means `true` and in this case setting `success` to 1 means the previous step failed, i.e. success is not true.

Also, could we not just wrap the retry in a while look that checks the retry counter instead of the success variable? Technically, success is a variable specific to a single attempt and shouldn't even be used outside the loop or as a condition for it. Instead, we should rely on the retry counter, i.e. `if attempt # < attempt limit, run retry`

srprash (approved):

PaurushGarg (approved):

PR #643, opened 2023-12-12
Provide aws-region for the e2e test in worklow

vasireddy99:
Current [workflows are failing](https://github.com/aws-observability/aws-otel-java-instrumentation/actions/runs/7176507664) and are not valid. Provide `aws-region` for the e2e test in worklow

### DIFF
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -461,6 +461,7 @@ jobs:
     uses: ./.github/workflows/appsignals-e2e-eks-test.yml
     secrets: inherit
     with:
+      aws-region: us-west-2
       test-cluster-name: "e2e-adot-test"
       appsignals-adot-image-name: ${{ needs.build.outputs.staging-image-name }}
       caller-workflow-name: 'main-build'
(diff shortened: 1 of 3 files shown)


.github/workflows/main-build.yml
```diff
@@ -461,6 +461,7 @@ jobs:
     uses: ./.github/workflows/appsignals-e2e-eks-test.yml
     secrets: inherit
     with:
+      aws-region: us-west-2
       test-cluster-name: "e2e-adot-test"
```
majanjua-amzn, line 465: The cluster `e2e-adot-test` does not exist in us-west-2. The region needs to be `us-east-1` and there needs to be a concurrency check to ensure that when this action is run the other workflows that use the same cluster do not run at the same time.

majanjua-amzn (changes requested):

.github/workflows/main-build.yml
bryan-aguilar: I believe this PR is to explicitly address the workflow failure that is occurring in the main build workflow

humivo (approved):

PR #652, opened 2023-12-14
E2E Testing: Fix EKS test candidate image override

majanjua-amzn:
This change checks if there is an adot image passed to the workflow and patches the App Signals deployment to update the image and restarts the cloudwatch pods. We check for input instead of a specific workflow, allowing any workflow to call this and run the e2e test.

Note: It is a limitation of EKS addon that we cannot set the artifact version before enabling pulse because the enablement script refers to a specific version of the addon we cannot change/create a new one for. As such, we need to apply the change for the ADOT image after enabling app signals, then restart the relevant pods.

Tested in private repo:
Test showing failure when using a non application signals image: [run](https://github.com/aws-observability/aws-apm-java-instrumentation/actions/runs/7201809370)
- `public.ecr.aws/aws-observability/adot-autoinstrumentation-java:v1.30.0`

Test showing pass for no input: [run](https://github.com/aws-observability/aws-apm-java-instrumentation/actions/runs/7212987747)
Test showing pass for app signals as input: [run](https://github.com/aws-observability/aws-apm-java-instrumentation/actions/runs/7213496367)
- `public.ecr.aws/aws-observability/adot-autoinstrumentation-java:v1.31.1`

### DIFF
--- a/.github/workflows/appsignals-e2e-eks-test.yml
+++ b/.github/workflows/appsignals-e2e-eks-test.yml
@@ -88,10 +88,6 @@ jobs:
         working-directory: testing/terraform/eks
         run: aws s3 cp ${{ env.ENABLEMENT_SCRIPT_S3_BUCKET }} . && unzip -j onboarding.zip
 
-      - name: Change ADOT image if main-build
-        if: inputs.caller-workflow-name == 'main-build'
-        run: "sed -i 's#image:.*#image: ${{ inputs.appsignals-adot-image-name }}#g' instrumentation.yaml"
-
       - name: Remove log group deletion command
         if: always()
         working-directory: testing/terraform/eks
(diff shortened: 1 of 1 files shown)


.github/workflows/appsignals-e2e-eks-test.yml
```diff
@@ -140,6 +136,15 @@ jobs:
               ${{ env.AWS_DEFAULT_REGION }} \
               ${{ env.SAMPLE_APP_NAMESPACE }}
           
+              # If the workflow provides a specific ADOT image to test, patch the deployment and restart CW agent related pods
+              if [ ${{ inputs.appsignals-adot-image-name }} != "" ]; then
+                kubectl patch deploy -namazon-cloudwatch amazon-cloudwatch-observability-controller-manager --type='json' \
+                -p='[{"op": "replace", "path": "/spec/template/spec/containers/0/args/0", "value": "--auto-instrumentation-java-image=${{ inputs.appsignals-adot-image-name }}"}]'
```
srprash, line 142: Can you provide link to the template code in CWAgent for this value that is being patched? Is it always guaranteed to be this path?

majanjua-amzn: Yes, the image for ADOT and for the Operator are set in this file:
https://github.com/aws/amazon-cloudwatch-agent-operator/blob/main/helm/templates/operator-deployment.yaml

srprash (approved):
LGTM

PaurushGarg (approved):

PR #627, opened 2023-12-04
nightly-build workfow update

PaurushGarg:
*Issue #, if available:*

*Description of changes:*

- [x] Pull Contract Tests in another job
- [x] Add timestamp tag to nightlybuild image
- [x] Add non-appsignals tests to nightlybuild workflow (included with agent and without agent in sample app e2e tests).
- [x] Pull common tests between main and nigthly build in separate workflows.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #689, opened 2024-01-03
Updating workflows

PaurushGarg:
*Issue #, if available:*

*Description of changes:*
Fixing this PR: https://github.com/aws-observability/aws-otel-java-instrumentation/pull/627
1. Separate out contract tests for main and nightly build.
2. Remove patch build for nightlybuild workflow as this workflow should need patching.


.github/workflows/nightly-upstream-snapshot-build.yml
```diff
@@ -45,22 +45,6 @@ jobs:
...
-      # cache local patch outputs
-      - name: Cache local Maven repository
-        id: cache-local-maven-repo
-        uses: actions/cache@v3
-        with:
-          path: |
-            ~/.m2/repository/io/opentelemetry/
-          key: ${{ runner.os }}-maven-local-${{ hashFiles('.github/patches/opentelemetry-java*.patch') }}
-
-      - name: Publish patched dependencies to maven local
```
vasireddy99, line 57: the reason for removing is that patching is not required in the nightly build workflows as it testing against upstream snapshot?

.github/workflows/nightly-upstream-snapshot-build.yml
```diff
@@ -115,24 +99,38 @@ jobs:
       test_ref: 'terraform'
       caller-workflow-name: 'nightly-upstream-snapshot-build'
 
-  # E2E tests where SampleApp has Java Agent
```
PaurushGarg, line 118: Removed this Test as it builds the sample app with the released upstream dependencies and not with the snapshot one. I think providing this test here, requires much more work to ensure sample apps are build with java agent that have upstream snapshot versions.

bryan-aguilar: Was this job recently added to the nightly build workflow? I am having a bit of trouble tracing the git blame back for it.

PaurushGarg: Yes I added it recently in [this](https://github.com/aws-observability/aws-otel-java-instrumentation/pull/627) PR. and I realized it was a mistake, so removing it with this PR. 
I think this testing should be part of nightly build, but putting it in `nightlybuild` workflows require some updates in the repo to facilitate this.

rapphil (approved):

PR #690, opened 2024-01-05
Dry run validator before pull request

XinRanZhAWS:
*Issue #, if available:*
we need a way to confirm validator (for test) is still be able to build after every code update

*Description of changes:*
Allow to dry run validator before pull request 


.github/workflows/pr-build.yml
```diff
@@ -80,6 +80,11 @@ jobs:
           gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
           gpg_password: ${{ secrets.GPG_PASSPHRASE }}
 
+      - name: Dry Run Validator
+        if: ${{ matrix.os != 'windows-latest' }} # Skip dry run validator on windows as it is not possible to build opentelemetry-java on windows
+        run: |
+          ./gradlew testing:validator:run --args='--dry-run-compile true'
```
rapphil, line 86: you don't need this to just compile. 

you can use `./gradlew testing:validator:build`

rapphil (changes requested):
Please consider taking a look into the gradle manual:  https://docs.gradle.org/current/userguide/tutorial_using_tasks.html

And also run
`./gradlew testing:validator:tasks --all`
so that you can get to know all the available tasks.

.github/workflows/pr-build.yml
XinRanZhAWS: Done with the change, thanks

XinRanZhAWS (commented):
Done with the change, thanks

.github/workflows/pr-build.yml
```diff
@@ -80,6 +80,11 @@ jobs:
           gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
           gpg_password: ${{ secrets.GPG_PASSPHRASE }}
 
+      - name: Dry Run Validator
+        if: ${{ matrix.os != 'windows-latest' }} # Skip dry run validator on windows as it is not possible to build opentelemetry-java on windows
+        run: |
+          ./gradlew testing:validator:build
+
```
bryan-aguilar, line 87: This should be in it's own job so it can run independently. Ideally this would only be ran when validator files change but I'm not sure how much time that would save us.

XinRanZhAWS: Since validator and the other code sharing same gradle file, this test added here aimed to avoid a case that the validator not change and the code repo changed to broke the validator.

bryan-aguilar: Can we use the gradle-build-action instead? That way we can utilize the caching by default.

bryan-aguilar (approved):

rapphil (approved):

PR #701, opened 2024-01-09
Build patched artifacts in codeql workflow

bryan-aguilar:
*Issue #, if available:* If a patch exists it must also be built in codeql workflow for successful workflow execution. 

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -25,15 +25,36 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v4
 
+    - uses: actions/setup-java@v4
+      with:
+        java-version: 17
+        distribution: temurin
+
+    - name: Cache local Maven repository
+      uses: actions/cache@v3
+      with:
+        path: |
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -25,15 +25,36 @@ jobs:
...
+
+    - name: Cache local Maven repository
+      uses: actions/cache@v3
+      with:
+        path: |
+          ~/.m2/repository/io/opentelemetry/
+        key: ${{ runner.os }}-maven-local-${{ hashFiles('.github/patches/opentelemetry-java*.patch') }}
+
+    - name: Publish patched dependencies to maven local
+      uses: ./.github/actions/patch-dependencies
```
PaurushGarg, line 41: @bryan-aguilar should we have a if condition here, like below?
`        if: steps.cache-local-maven-repo.outputs.cache-hit != 'true'
`

bryan-aguilar: added

PaurushGarg (approved):

PR #704, opened 2024-01-09
Set default region as build output

bryan-aguilar:
You cannot use env vars in the context of reusable workflow inputs. The env var must first be set as an output for a previous job. 


PaurushGarg (changes requested):
@bryan-aguilar we need to change aws-region with in main-build and nightly-build for other tests/jobs as well, like e2e-test, e2e-operator-test

bryan-aguilar:
@PaurushGarg done

PaurushGarg (approved):

PR #703, opened 2024-01-09
Fix usage of input variable in workflows

bryan-aguilar:
*Description of changes:* Refactor main build workflow to use env var directly. Also remove usage of setting DEFAULT_REGION env var in favor of just using input variable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #708, opened 2024-01-10
Reinsert aws default region env var

bryan-aguilar:
*Description of changes:* Revert the changes to `AWS_DEFAULT_REGION` env var that were made #703. The presence of the env var is required for some steps inside the workflows. This is not redundant and is required to be present in the env. 

This PR also moves the setting of the default region output to an explicit job. Previously it was done at the end of the build step. This caused `us-west-2` to be output because the composite `cpUtility` test action overrode that default value. Using an explicit job here will ensure that `us-east-1` value is used as the output every time. 

### DIFF
--- a/.github/workflows/appsignals-e2e-ec2-test.yml
+++ b/.github/workflows/appsignals-e2e-ec2-test.yml
@@ -17,6 +17,9 @@ permissions:
   contents: read
 
 env:
+  # The precense of this env var is required for use by terraform and AWS CLI commands
+  # It is not redundant
+  AWS_DEFAULT_REGION: ${{ inputs.aws-region }} 
   TEST_ACCOUNT: ${{ secrets.APP_SIGNALS_E2E_TEST_ACC }}
   SAMPLE_APP_FRONTEND_SERVICE_JAR: "s3://aws-appsignals-sample-app/main-service.jar"
   SAMPLE_APP_REMOTE_SERVICE_JAR: "s3://aws-appsignals-sample-app/remote-service.jar"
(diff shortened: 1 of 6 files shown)


PaurushGarg (commented):
@bryan-aguilar I am not getting it. Where are we using `AWS_DEFAULT_REGION` ?

bryan-aguilar:
See [this](https://github.com/aws-observability/aws-otel-java-instrumentation/actions/runs/7468855410/job/20325269013#step:8:17) error for an example. We do not explicitly set the region in that login action, so it uses `us-west-2`. The presence of `AWS_DEFAULT_REGION` is a picked up by the aws cli/sdk and is used.

bryan-aguilar:
Although, I just realized another error at the same time. I will adjust the PR. We need a separate job for exporting the default env var. It is getting overwritten.

PaurushGarg:
> See [this](https://github.com/aws-observability/aws-otel-java-instrumentation/actions/runs/7468855410/job/20325269013#step:8:17) error for an example. We do not explicitly set the region in that login action, so it uses `us-west-2`. The presence of `AWS_DEFAULT_REGION` is a picked up by the aws cli/sdk and is used.

ok. Thanks.
but for cpUtility test (step, we are passing us-west-2 as input aws-region, would that be a problem there? https://github.com/bryan-aguilar/aws-otel-java-instrumentation/blob/6f528014da848b05e95ab4624f43f727a63c272b/.github/workflows/main-build.yml#L125

bryan-aguilar:
Yeah it is. which is why we are ending up with us-west-2. I still think we need those env vars set though. So this will become a two part fix.

PaurushGarg (approved):

PR #712, opened 2024-01-11
Change terraform to use minimal ami

harrryr:
*Issue #, if available:*
A new AMI available to the public that was matching our filter conditions for EC2 Terraform came pre-installed with Java 8. This caused dependency issues with the sample app which uses Java 11 used for E2E testing and caused the canary to fail. 

*Description of changes:*
Added additional filter conditions for the EC2 AMIs to look for minimal images. Also updated the instances to install `java-11-amazon-coretto`

Test run:
us-east-1: https://github.com/harrryr/aws-otel-java-instrumentation/actions/runs/7495423753
us-east-2: https://github.com/harrryr/aws-otel-java-instrumentation/actions/runs/7495550565
eu-west-1: https://github.com/harrryr/aws-otel-java-instrumentation/actions/runs/7495547623
ap-norhteast-1: https://github.com/harrryr/aws-otel-java-instrumentation/actions/runs/7495545124
ap-southeast-2: https://github.com/harrryr/aws-otel-java-instrumentation/actions/runs/7495438492

### DIFF
--- a/testing/terraform/ec2/main.tf
+++ b/testing/terraform/ec2/main.tf
@@ -89,10 +89,8 @@ resource "null_resource" "main_service_setup" {
 
   provisioner "remote-exec" {
     inline = [
-      # Install wget
-      "yes | sudo yum install wget",
-      # Install Java 11
-      "sudo yum install java-11-amazon-corretto -y",
+      # Install Java 11 and wget
+      "sudo yum install wget java-11-amazon-corretto -y",
 
       # Copy in CW Agent configuration
       "agent_config='${replace(replace(file("./amazon-cloudwatch-agent.json"), "/\\s+/", ""), "$REGION", var.aws_region)}'",
@@ -152,10 +150,8 @@ resource "null_resource" "remote_service_setup" {
 
   provisioner "remote-exec" {
     inline = [
-      # Install wget
-      "yes | sudo yum install wget",
-      # Install Java 11
-      "sudo yum install java-11-amazon-corretto -y",
+      # Install Java 11 and wget
+      "sudo yum install wget java-11-amazon-corretto -y",
 
       # Copy in CW Agent configuration
       "agent_config='${replace(replace(file("./amazon-cloudwatch-agent.json"), "/\\s+/", ""), "$REGION", var.aws_region)}'",
--- a/testing/terraform/ec2/main.tf
+++ b/testing/terraform/ec2/main.tf
@@ -27,8 +27,16 @@ locals {
 }
 
 data "aws_ami" "ami" {
-  executable_users = ["self"]
+  owners = ["amazon"]
   most_recent      = true
+  filter {
+    name   = "name"
+    values = ["al20*-ami-minimal-*-x86_64"]
+  }
+  filter {
+    name   = "state"
+    values = ["available"]
+  }
   filter {
     name   = "architecture"
     values = ["x86_64"]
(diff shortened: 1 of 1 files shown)


testing/terraform/ec2/main.tf
```diff
@@ -81,8 +89,10 @@ resource "null_resource" "main_service_setup" {
 
   provisioner "remote-exec" {
     inline = [
-      # Install Java 11 and tmux
-      "yes | sudo amazon-linux-extras install java-openjdk11",
+      # Install wget
+      "yes | sudo yum install wget",
+      # Install Java 11
+      "sudo yum install java-11-amazon-corretto -y",
```
pxaws, line 95: minor: you can install both in one line

```
sudo yum install wget java-11-amazon-corretto -y
```

testing/terraform/ec2/main.tf
```diff
@@ -27,8 +27,16 @@ locals {
 }
 
 data "aws_ami" "ami" {
-  executable_users = ["self"]
+  owners = ["amazon"]
   most_recent      = true
+  filter {
+    name   = "name"
+    values = ["al20*-ami-minimal-*-x86_64"]
```
srprash, line 34: NB comment: Do we need to specify `x86_64` in the name even though we have additional filter on the architecture down below?

srprash (approved):

testing/terraform/ec2/main.tf
majanjua-amzn: Also, how have we verified that the AMI identified here will always be usable?

pxaws (approved):

testing/terraform/ec2/main.tf
harrryr: Done and tested here: https://github.com/harrryr/aws-otel-java-instrumentation/actions/runs/7495714707

testing/terraform/ec2/main.tf
```diff
@@ -142,8 +150,10 @@ resource "null_resource" "remote_service_setup" {
 
   provisioner "remote-exec" {
     inline = [
-      # Install Java 11 and tmux
-      "yes | sudo amazon-linux-extras install java-openjdk11",
+      # Install wget
+      "yes | sudo yum install wget",
+      # Install Java 11
+      "sudo yum install java-11-amazon-corretto -y",
```
majanjua-amzn, line 156: Make the same change Ping commented about here as well, and then rerun the test

testing/terraform/ec2/main.tf
harrryr: Prashant: I believe it's redundant, is this something we prefer to have removed?
Mahad: There is a depreciation date for each ami, usually around ~3 month. It looks like a new version of this minimal AMI is released every ~2 weeks (Period is kind of infrequent) and I can see it going back to Oct 2023. (Further back was probably depreciated so I can't see). I think this AMI has been being maintained for a while and we can expect newer versions to keep getting released.

pxaws: The AMI name is defined by convention. I find one public doc talked about it [here](https://docs.aws.amazon.com/linux/al2023/ug/ec2.html#launch-from-ec2-console)

>al2023-[ami || ami-minimal]-2023.0.[release build date].[build number]-kernel-[version number]-[arm64 || x86_64]

I think we can mostly assume the convention will be kept. Though no perfect solution here.

bryan-aguilar (approved):

PR #729, opened 2024-01-25
Add tests for Metric Rollup Addition of [RemoteService, RemoteTarget, ...]

jj22ee:
*Issue #, if available:*
https://github.com/aws/amazon-cloudwatch-agent/pull/1010
Recently the following metric rollups were added to 2 AppSignal Configuration files. Tests need to be added for them
  - `[RemoteService, RemoteTarget]`
  - `[HostedIn.<Attributes>, Service, RemoteService, RemoteTarget]`

*Description of changes:*
Add tests for new metric rollups

*Testing done:*
EKS in IAD: https://github.com/jj22ee/amazon-cloudwatch-agent/actions/runs/7659009903/job/20873170967
EC2 in IAD: https://github.com/jj22ee/amazon-cloudwatch-agent/actions/runs/7659098851/job/20873465899

`./gradlew testing:validator:test`

### DIFF
--- a/testing/validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
+++ b/testing/validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
@@ -15,16 +15,6 @@
 
 package com.amazon.aoc.validators;
 
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
-import java.util.TreeSet;
-
 import com.amazon.aoc.callers.ICaller;
 import com.amazon.aoc.exception.BaseException;
 import com.amazon.aoc.exception.ExceptionCode;
@@ -38,7 +28,15 @@ import com.amazon.aoc.services.CloudWatchService;
 import com.amazonaws.services.cloudwatch.model.Dimension;
 import com.amazonaws.services.cloudwatch.model.Metric;
 import com.google.common.collect.Lists;
-
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.TreeSet;
 import kotlin.Pair;
 import lombok.extern.log4j.Log4j2;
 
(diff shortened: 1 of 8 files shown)


testing/validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
```diff
@@ -103,16 +105,27 @@ public void validate() throws Exception {
...
+            dimensionLists.add(
+                Arrays.asList(
+                    new Pair<>(CloudWatchService.REMOTE_SERVICE_DIMENSION, remoteServiceName)));
+          }
+          dimensionLists.add(
+              Arrays.asList(
+                  new Pair<>(CloudWatchService.REMOTE_SERVICE_DIMENSION, "AWS.SDK.S3"),
+                  new Pair<>(CloudWatchService.REMOTE_TARGET_DIMENSION, "e2e-test-bucket-name")));
+
+          // Populate actualMetricList with each set of dimension filters
```
majanjua-amzn, line 125:
```suggestion
          // Populate actualMetricList with metrics that pass through one of the dimension filters
```

testing/validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
```diff
@@ -103,16 +105,27 @@ public void validate() throws Exception {
...
-              serviceNames,
-              expectedMetricList,
-              actualMetricList);
-          addMetrics(
-              CloudWatchService.REMOTE_SERVICE_DIMENSION,
-              remoteServiceNames,
-              expectedMetricList,
-              actualMetricList);
+
+          // Add sets of dimesion filters to use for each query to CloudWatch
```
majanjua-amzn, line 109: We've now added a second edge case to this logic, i.e. the [RemoteService] and [RemoteService, RemoteTarget] aggregations. It would be worth adding a larger comment here explaining what's happening and explain why in the PR description for future reference.

testing/validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
```diff
@@ -103,16 +105,27 @@ public void validate() throws Exception {
...
+          List<List<Pair<String, String>>> dimensionLists = Lists.newArrayList();
+          for (String serviceName : serviceNames) {
+            dimensionLists.add(
+                Arrays.asList(new Pair<>(CloudWatchService.SERVICE_DIMENSION, serviceName)));
+          }
+          for (String remoteServiceName : remoteServiceNames) {
+            dimensionLists.add(
+                Arrays.asList(
+                    new Pair<>(CloudWatchService.REMOTE_SERVICE_DIMENSION, remoteServiceName)));
+          }
```
majanjua-amzn, line 119: Is this new logic able to pull `[RemoteService, RemoteTarget]` without pulling `[<other stuff>, RemoteService, RemoteTarget]`? If yes, we can add back the other remote service values to `remoteServiceNames`. If not, the code changes in this PR will cause flaky tests in the long run.

This is due to the fact that when trying to pull `[RemoteService, RemoteTarget]` if you also pull other aggregations that have the same two attributes you will pull metrics from all test runs in the past three hours that match the filter you set. CW listMetrics API cannot return all of these metrics reliably and we've seen this cause failures before. Please ensure your solution accounts for this by running a few tests in a row and seeing if metrics with attributes containing the wrong test ID appear in the `actualMetricList`.

testing/validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
```diff
@@ -103,16 +105,27 @@ public void validate() throws Exception {
...
+          dimensionLists.add(
+              Arrays.asList(
+                  new Pair<>(CloudWatchService.REMOTE_SERVICE_DIMENSION, "AWS.SDK.S3"),
+                  new Pair<>(CloudWatchService.REMOTE_TARGET_DIMENSION, "e2e-test-bucket-name")));
```
majanjua-amzn, line 123: Once the metric exists, how do we know whether the test that just occurred is the one that populated it or if it simply exists from a previous test run? It's entirely possible that a test from 2 hours ago created the metric and then all tests since then did not appropriately populate this metric but because of our `.withRecentlyActive("PT3H")` we are not finding out at all that this metric is flakily populated.

This is the same issue we had with trying to check `[RemoteService]` when it is equal to `["www.amazon.com"]` for example. (There are also other considerations for this aggregation which I'll mention in another comment)

Note: `withRecentlyActive()` only accepts PT3H

majanjua-amzn (changes requested):
Overall seems to be working but there's some design questions and investigations into the unintended effects of the way the logic works to check.

Also, do not merge this code until CW Agent performs their release, otherwise you will cause canary to break leading to sev 2.

testing/validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
majanjua-amzn: Please be more descriptive in other comments in your PR as well.

jj22ee:
In discussion, a solution for both issues (`[RemoteService, RemoteTarget]` pulling in too much data which comes from previous test runs and can validate the current test run) is to update the aws-sdk-call api to take in an identifier (like `Pod_IP`) and use that in the bucket name (e.g. `e2e-test-bucket-name-<Pod_IP>`). Then the tests will need to specify this ID in queries that include RemoteTarget dimension.

jj22ee:
Closing in favor of: https://github.com/aws-observability/aws-otel-python-instrumentation/pull/164

PR #745, opened 2024-02-09
Remote target

atshaw43:
*Description of changes:*
Setting remote target to ARN for SQS non-custom URLs. Set remote target to ARN like structure for SQS customer URLs, Kinesis, DynamoDB, and S3.


awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -144,13 +135,27 @@ private static void setRemoteTarget(SpanData span, AttributesBuilder builder) {
...
+      String arn = SqsUrlParser.getSqsRemoteTarget(span.getAttributes().get(AWS_QUEUE_URL));
+
+      if (arn != null) {
+        return Optional.ofNullable(arn);
+      }
+    }
+
+    if (isKeyPresent(span, AWS_QUEUE_NAME)) {
+      return Optional.ofNullable("::sqs:::" + span.getAttributes().get(AWS_QUEUE_NAME));
+    }
```
mxiamxia, line 151: are these two redundant for SQS?  Will we get both SQS ARN and SQS queue name?

atshaw43: We cannot get both. If the first succeeds, then it exits. We try to use the URL first because it has more data potentially.

mxiamxia (approved):

bryan-aguilar (approved):

PR #751, opened 2024-02-16
Adding support for ISO regions

atshaw43:
*Description of changes:*
Adding support for ISO regions in the remote target for SQS.
Added additional validation for region and account id.
Fixed some tests. The testing logic that already existed was busted because if the target compared to was null it didn't validate it.


mxiamxia (approved):

thpierce:
No longer pursuing this.

PR #755, opened 2024-02-21
Parse db.statement attribute for database instrumentation

AsakerMohd:
Description of changes:
When `"db.operation"` is not provided in span attributes, the `"aws.remote.operation"` will be `UnknowRemoteOperation`.
This PR updated the `setRemoteServiceAndOperation` function to parsing the `"db.statement"` attribute when `"db.operation"` is missing, and extract the valid `"aws.remote.operation"` (We setup a set of valid database related `"aws.remote.operation"` for security concern):
1. Add a json file `configuration/dialect_keywords.json` saving all the valid keywords. In this file, the sequence of key words matter -> The keyword with longer word length are placed towards the front of the list so it will be matched first. 
2. We retrieve the first 27 characters from `"db.statement"` to avoid the case where very large statements, and use regular expression to match the keyword, it will match the beginning of the string. If the string's start does not conform to the regular expression, the match fails.
3. Add unit test for `setRemoteServiceAndOperation` performance where tested DB_STATEMENT is/is not present and is/is not valid.
4. Add unit test covering different `"db.statement"` cases:
    1. Only 1 valid keywords match
    1. More than 1 valid keywords match, we want to pick the longest match
    1. More than 1 valid keywords match, but the other keywords is not at the start of the SpanAttributes.DB_STATEMENT. We want to only pick start match
    1. No valid match
    1. Have valid but it is not at the start of SpanAttributes.DB_STATEMENT
    1. Have valid keywords, match the longest word
    1. Have valid keywords, match with first word
    1. Have valid keywords, match with upper case
    1. Have both `"db.statement"` and `"db.operation"` set and the former is returned
5. Add unit test for testing keywords sequence in json file:
    1. Confirm the keywords are sorted based on descending order of keywords character length
    1. Confirm maximum length of keywords is not longer than MAX_KEYWORD_LENGTH

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.


awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -449,6 +460,43 @@ private static String getRemoteOperation(SpanData span, AttributeKey<String> rem
...
+    // Remove all whitespace and newline characters from the beginning of remote_operation
+    // and retrieve the first MAX_KEYWORD_LENGTH characters
+    remoteOperation = remoteOperation.stripLeading();
+    if (remoteOperation.length() > MAX_KEYWORD_LENGTH) {
+      remoteOperation = remoteOperation.substring(0, MAX_KEYWORD_LENGTH);
+    }
+
+    try {
+      String joinedKeywords = String.join("|", getDialectKeywords());
+      Pattern pattern = Pattern.compile("^(?:" + joinedKeywords + ")\\b");
```
srprash, line 485: Why do we need to get the keywords, join then, and compile the pattern every time? It's not efficient. We can just make this a constant in the `AwsSpanProcessingUtil` class. for example, we can do:
```java
static final Pattern DB_STATEMENT_KEYWORDS_PATTERN = Pattern.compile("^(?:" + String.join("|", getDialectKeywords()) + ")\\b");
```

And then do `Matcher matcher = DB_STATEMENT_KEYWORDS_PATTERN.matcher(remoteOperation.toUpperCase());` below.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsSpanProcessingUtil.java
```diff
@@ -40,6 +47,19 @@ final class AwsSpanProcessingUtil {
   static final String LOCAL_ROOT = "LOCAL_ROOT";
   static final String SQS_RECEIVE_MESSAGE_SPAN_NAME = "Sqs.ReceiveMessage";
   static final String AWS_SDK_INSTRUMENTATION_SCOPE_PREFIX = "io.opentelemetry.aws-sdk-";
+  // Max keyword length supported by parsing into remote_operation from DB_STATEMENT
```
srprash, line 50: Can you also describe in the comment how we arrived at the value of the `MAX_KEYWORD_LENGTH` and if we ever need to update this value, how should we decide the new value?

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsSpanProcessingUtil.java
```diff
@@ -40,6 +47,19 @@ final class AwsSpanProcessingUtil {
...
+
+  static List<String> getDialectKeywords() throws IOException {
+    InputStream jsonFile =
+        AwsSpanProcessingUtil.class.getClassLoader().getResourceAsStream(SQL_DIALECT_KEYWORDS_JSON);
+    ObjectMapper mapper = new ObjectMapper();
+    JsonNode jsonNode = mapper.readValue(jsonFile, JsonNode.class);
+    JsonNode arrayNode = jsonNode.get("keywords");
+    ObjectReader reader = mapper.readerFor(new TypeReference<List<String>>() {});
+    return reader.readValue(arrayNode);
+  }
```
srprash, line 62: Related to my comment about using the json file and reading it, this logic could be much simpler if we are using a native datatype (`Set` maybe?) to store the db statement keywords.

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
```diff
@@ -461,6 +462,21 @@ public void testRemoteAttributesCombinations() {
     // Validate behaviour of various combinations of DB attributes, then remove them.
     validateAndRemoveRemoteAttributes(DB_SYSTEM, "DB system", DB_OPERATION, "DB operation");
 
+    // Validate db.operation not exist, but db.statement exist, where SpanAttributes.DB_STATEMENT is
```
srprash, line 465: Can we also add a test case where both `db.operation` and `db.statement` exists but we pick the value from the former?

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsSpanProcessingUtil.java
srprash: Oh. just realized that the keywords need to be ordered, so can't use `Set`. A simple `List` then, although it will not deduplicate the elements.

srprash:
The PR description doesn't correctly mention the actual java methods used in this PR. Please fix the description to be more accurate.

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
AsakerMohd: Added test case 9.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsSpanProcessingUtil.java
AsakerMohd: The keywords in the list are supposed to be deduplicated already.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsSpanProcessingUtil.java
AsakerMohd: Added comment describing why we choose 27.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
AsakerMohd: Yeah that is more efficient for sure. Updated.

awsagentprovider/build.gradle.kts
```diff
@@ -34,6 +34,8 @@ dependencies {
   implementation("io.opentelemetry.contrib:opentelemetry-aws-xray")
   // AWS Resource Detectors
   implementation("io.opentelemetry.contrib:opentelemetry-aws-resources")
+  // Json file reader
+  implementation(group = "com.fasterxml.jackson.dataformat", name = "jackson-dataformat-yaml", version = "2.15.1")
```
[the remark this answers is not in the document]
AsakerMohd: Updated the dependency. As for storing in file vs in class, I'm not sure that matters much since now it it will read the file only once when the class is first initialized/called and after that, the list is stored in memory anyway.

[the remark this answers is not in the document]
srprash: Ok. I'm fine with reading the keywords from the file for now.

srprash (approved):

vasireddy99 (approved):

PR #756, opened 2024-02-22
prefix `AWS.SDK` to the RemoteService from AWS SDK v1 spans

srprash:
A follow up on this PR, where we left out the AWS SDK v1 instrumentation.
https://github.com/aws-observability/aws-apm-java-instrumentation/pull/160

Now with this change, for any version of AWS SDK client, we will be prefixing `AWS.SDK.` to the `RemoteService` attribute for the generated AppSignals metrics.

### DIFF
--- a/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
+++ b/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
@@ -216,10 +216,7 @@ final class AwsMetricAttributeGenerator implements MetricAttributeGenerator {
   // add `AWS.SDK.` as prefix to indicate the metrics resulted from current span is from AWS SDK
   private static String normalizeServiceName(SpanData span, String serviceName) {
     if (AwsSpanProcessingUtil.isAwsSDKSpan(span)) {
-      String scopeName = span.getInstrumentationScopeInfo().getName();
-      if (scopeName.contains("aws-sdk-2.")) {
-        return "AWS.SDK." + serviceName;
-      }
+      return "AWS.SDK." + serviceName;
     }
     return serviceName;
   }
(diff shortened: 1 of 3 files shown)


awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -216,10 +216,7 @@ private static void setEgressOperation(SpanData span, AttributesBuilder builder)
   // add `AWS.SDK.` as prefix to indicate the metrics resulted from current span is from AWS SDK
   private static String normalizeServiceName(SpanData span, String serviceName) {
     if (AwsSpanProcessingUtil.isAwsSDKSpan(span)) {
-      String scopeName = span.getInstrumentationScopeInfo().getName();
-      if (scopeName.contains("aws-sdk-2.")) {
-        return "AWS.SDK." + serviceName;
-      }
+      return "AWS.SDK." + serviceName;
```
mxiamxia, line 219: Do we actually need `InstrumentationScopeName` as our second gatekeeper to confirm the mutated span name is AWS service? Or we're pretty confident that we can solely rely on `isAwsSDKSpan`?

srprash: The [semantic convention for AWS SDK](https://opentelemetry.io/docs/specs/semconv/cloud-providers/aws-sdk/) requires the `rpc.system` attribute value to be `aws-api` and [that's what the `isAwsSDKSpan` method checks for](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/7fdbf2185facf4a1d2dbbd334c4e44262de5362e/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsSpanProcessingUtil.java#L120). 

So I think we can rely on the `isAwsSDKSpan` alone here. If for some reason, the AWS SDK instrumentation stops populating that attribute, then the contract tests will catch it.

mxiamxia (approved):

vasireddy99 (approved):

PR #838, opened 2024-07-10
Clean up E2E Operator resources if tests fail or are cancelled

thpierce:
In this PR, we are changing E2E Operator tests to clean up after themselves on failure or cancellation. Currently, this only happens on cancellation, but we need to do this on failure too, as tests can fail due to leaked resources.

This operation is safe to perform, as E2E Operator tests are only run from [two locations](https://github.com/search?q=repo%3Aaws-observability%2Faws-otel-java-instrumentation%20e2e-tests-with-operator.yml&type=code):
1. [main-build](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/90901b191b0f9e934a4c698ec1d1448364ad86fd/.github/workflows/main-build.yml#L160-L162)
2. [nightly-upstream-snapshot-build](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/90901b191b0f9e934a4c698ec1d1448364ad86fd/.github/workflows/nightly-upstream-snapshot-build.yml#L99-L102)

In both cases, they are running with a concurrency group of `e2e-adot-agent-operator-test` which prevents overlapping test runs.

Note: A number of code samples showing `cancelled or failed`: https://github.com/search?q=%22if%3A+%24%7B%7B+cancelled%28%29+%7C%7C+failure%28%29+%7D%7D%22&type=code

### DIFF
--- a/.github/workflows/e2e-tests-with-operator.yml
+++ b/.github/workflows/e2e-tests-with-operator.yml
@@ -189,7 +189,7 @@ jobs:
       # file locks are being held from SIGTERMS dispatched in previous
       # steps.
       - name: Destroy resources
-        if: ${{ cancelled() }}
+        if: ${{ cancelled() || failure() }}
         shell: bash {0}
         working-directory: testing-framework/terraform
         run: |


.github/workflows/e2e-tests-with-operator.yml
```diff
@@ -209,4 +209,4 @@ jobs:
       success: ${{  needs.run-batch-job.result == 'success' }}
       region: us-west-2
     secrets:
-      roleArn: ${{ secrets.METRICS_ROLE_ARN }}
\ No newline at end of file
+      roleArn: ${{ secrets.METRICS_ROLE_ARN }}
```
thpierce, line 212: I don't know why this happened, I used the Git editor and it just added it automatically.

Aneurysm9: The previous revision had no `\n` at `EOF`, which makes it an invalid text file (per the POSIX spec, most things don't really care).  That's indicated by the line-in-circle symbol in the diff.  The editor is doing the right thing.

Aneurysm9 (approved):

PR #839, opened 2024-07-10
Add opentelemetry-java-instrumentation.patch.

zzhlogin:
Pulls in the following commits:
commit 6f34ea37729460c3649aa3e6cfafc45a63298b65([zzhlogin:bedrock-patch](https://github.com/zzhlogin/aws-otel-java-instrumentation-xiamin/tree/bedrock-patch)) Add opentelemetry-java-instrumentation.patch.

Using command `git cherry-pick cherry-pick cdbd978d05baee4489f93abce1d7d7249f93abb2`

Change: Add patch for opentelemetry-java-instrumentation pkg with bedrock services support.


dependencyManagement/build.gradle.kts
```diff
@@ -27,7 +27,7 @@ data class DependencySet(val group: String, val version: String, val modules: Li
 val TEST_SNAPSHOTS = rootProject.findProperty("testUpstreamSnapshots") == "true"
 
 // This is the version of the upstream instrumentation BOM
-val otelVersion = "1.32.1"
+val otelVersion = "1.32.1-adot1"
```
bjrara, line 30: Sorry, why do we need this change?

bjrara (approved):

dependencyManagement/build.gradle.kts
zzhlogin: We need to update the otel dependency to use version `1.32.1-adot1`, so that it can use our patch agent instead of upstream default agent. 
We also override the version in patch file, it can be found at the end of .github/patches/opentelemetry-java-instrumentation.patch:
```
--- a/version.gradle.kts
+++ b/version.gradle.kts
@@ -1,5 +1,5 @@
-val stableVersion = "1.32.1"
-val alphaVersion = "1.32.1-alpha"
+val stableVersion = "1.32.1-adot1"
+val alphaVersion = "1.32.1-adot1-alpha"
```

vasireddy99 (approved):

PR #843, opened 2024-07-15
Support to add `-r` to cp-utility tool

vasireddy99:
*Description of changes:*

This PR adds support for cp -r  command to copy recursively, and recursive copy will handle symbolic links as a normal test file.

Unit tests added.

and tested using the local built binary

`./target/debug/cp-utility -r README.md Find.md  `

### DIFF
--- a/tools/cp-utility/src/main.rs
+++ b/tools/cp-utility/src/main.rs
@@ -52,19 +52,15 @@ fn parse_args(args: Vec<&str>) -> io::Result<CopyOperation> {
     }
 
     if args.len() == 4 {
-        if args[1].eq("-a") {
-            return Ok(CopyOperation {
-                source: PathBuf::from(args[2]),
-                destination: PathBuf::from(args[3]),
-                copy_type: CopyType::Archive,
-            });
-        } else if args[1].eq("-r") {
-            return Ok(CopyOperation {
-                source: PathBuf::from(args[2]),
-                destination: PathBuf::from(args[3]),
-                copy_type: CopyType::Recursive,
-            });
-        }
+        return Ok(CopyOperation {
+            source: PathBuf::from(args[2]),
+            destination: PathBuf::from(args[3]),
+            copy_type: match args[1] {
+                "-a" => CopyType::Archive,
+                "-r" => CopyType::Recursive,
+                _ => panic!("Invalid option. Expected -a or -r"),
+            },
+        });
     }
 
     Ok(CopyOperation {
--- a/tools/cp-utility/src/main.rs
+++ b/tools/cp-utility/src/main.rs
@@ -28,6 +28,8 @@ enum CopyType {
     SingleFile,
     /// equivalent to cp -a <source> <dest>
     Archive,
+    /// equivalent to cp -r <source> <dest>
+    Recursive,
 }
 
 /// Encapsulate a copy operation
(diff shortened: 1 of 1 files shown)


tools/cp-utility/src/main.rs
```diff
@@ -42,21 +44,32 @@ struct CopyOperation {
...
 fn parse_args(args: Vec<&str>) -> io::Result<CopyOperation> {
-    if !(args.len() == 3 || args.len() == 4 && args[1].eq("-a")) {
+    if !(args.len() == 3
+        || args.len() == 4 && args[1].eq("-a")
+        || args.len() == 4 && args[1].eq("-r"))
+    {
         return Err(io::Error::new(
             io::ErrorKind::InvalidInput,
-            "Invalid parameters. Expected cp [-a] <source> <destination>",
+            "Invalid parameters. Expected cp [-a] <source> <destination> or  Expected cp [-r] <source> <destination>",
```
rohang98, line 53: nit: Simplify to:

```
Expected cp [-a] <source> <destination> or cp [-r] <source> <destination>
```

tools/cp-utility/src/main.rs
```diff
@@ -42,21 +44,32 @@ struct CopyOperation {
...
         ));
     }
 
-    if args.len() == 4 {
+    if args.len() == 4 && args[1].eq("-a") {
```
rohang98, line 57: nit: Should `args[1].eq("-a")` be nested so `args[1].eq("-r")` can be nested as well?

tools/cp-utility/src/main.rs
Aneurysm9: Go even further:

```suggestion
            "Invalid parameters. Expected cp [-a | -r] <source> <destination>",
```

tools/cp-utility/src/main.rs
```diff
@@ -42,21 +44,32 @@ struct CopyOperation {
 
 /// Parse command line arguments and transform into `CopyOperation`
 fn parse_args(args: Vec<&str>) -> io::Result<CopyOperation> {
-    if !(args.len() == 3 || args.len() == 4 && args[1].eq("-a")) {
+    if !(args.len() == 3
+        || args.len() == 4 && args[1].eq("-a")
+        || args.len() == 4 && args[1].eq("-r"))
+    {
```
Aneurysm9, line 50: Is the associativity correct here or are more parens needed to make sure the logic is correct?  I don't see a test for this function, maybe that's needed.  Could probably also be simplified:

```suggestion
    if !(args.len() == 3
        || (args.len() == 4 && (args[1].eq("-a") || args[1].eq("-r")))
    {
```

tools/cp-utility/src/main.rs
```diff
@@ -69,10 +82,51 @@ fn do_copy(operation: CopyOperation) -> io::Result<()> {
     match operation.copy_type {
         CopyType::Archive => copy_archive(&operation.source, &operation.destination)?,
         CopyType::SingleFile => fs::copy(&operation.source, &operation.destination).map(|_| ())?,
+        CopyType::Recursive => copy_recursive(&operation.source, &operation.destination)?,
     };
     Ok(())
 }
 
+fn copy_recursive(source: &Path, dest: &Path) -> io::Result<()> {
```
Aneurysm9, line 90: What is the actual difference between this and `copy_archive`?  Could/should these be a single function with a parameter gating the difference?

vasireddy99: The main difference between these two functions is handling of the symbolic links. In `copy_recursive` it does not follow them by default https://www.gnu.org/software/coreutils/manual/html_node/cp-invocation.html and treat it as normal file.

on more difference that is currently not present/implemented between these two, is that copy_archive should store attributes of the file, where the copy_recursive will not. right now both copy_archive and copy_recursive are handling in same way without keeping any attributes of file. I am tending to keep these as seperate functions for copy_recursive and copy_archive function as it seems more clean to implement and unit tests . so copy_archive can be updated in future.

tools/cp-utility/src/main.rs
vasireddy99: yes yes, simplified the associativity and added parser_correct unit test

tools/cp-utility/src/main.rs
```diff
@@ -42,19 +44,27 @@ struct CopyOperation {
...
+                destination: PathBuf::from(args[3]),
+                copy_type: CopyType::Archive,
+            });
+        } else if args[1].eq("-r") {
+            return Ok(CopyOperation {
+                source: PathBuf::from(args[2]),
+                destination: PathBuf::from(args[3]),
+                copy_type: CopyType::Recursive,
+            });
+        }
```
Aneurysm9, line 67: Since these operations only differ by the `copy_type` field this can be simplified with a `match` statement.  (double check my syntax here)

```suggestion
           return Ok(CopyOperation {
              source: PathBuf::from(args[2]),
              destination: PathBuf::from(args[3]),
              copy_type: match args[1] {
                "-a" => CopyType::Archive,
                "-r" => CopyType::Recursive,
                _ => panic!("invalid argument")
                }
          });
```

vasireddy99: agreed, committed the changes

Aneurysm9 (approved):

PR #847, opened 2024-07-18
temp: Comment out ECR-related release steps

majanjua-amzn:
The 1.32.3 release failed to publish non-K8s/EKS related artifacts. As such, we will temporarily comment these out while working to resolve the authentication issues for the Maven artifacts that are not being published. In this way, we will only perform the failed steps and their prerequisite steps on the next attempt, so as to not change the image digest of the ECRs that have already been published.

Changing a github workflow file will NOT impact the released artifacts.

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -35,74 +35,74 @@ jobs:
           gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
           gpg_password: ${{ secrets.GPG_PASSPHRASE }}
 
-      - name: Configure AWS Credentials
-        uses: aws-actions/configure-aws-credentials@v4
-        with:
-          role-to-assume: ${{ secrets.AWS_ASSUME_ROLE_ARN }}
-          aws-region: ${{ env.AWS_PUBLIC_ECR_REGION }}
+      # - name: Configure AWS Credentials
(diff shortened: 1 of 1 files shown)


.github/workflows/release-build.yml
```diff
@@ -35,74 +35,74 @@ jobs:
...
+
+      # - name: Log in to AWS ECR
+      #   uses: docker/login-action@v3
+      #   with:
+      #     registry: public.ecr.aws
+
+      # - name: Build release with Gradle
+      #   uses: gradle/gradle-build-action@v3
+      #   with:
+      #     arguments: build integrationTests -PlocalDocker=true -Prelease.version=${{ github.event.inputs.version }} --stacktrace
```
srprash, line 52: We don't need this?

majanjua-amzn: Not needed, it builds docker images which is already done:
https://github.com/aws-observability/aws-otel-java-instrumentation/actions/runs/9995276354/job/27627037143#step:8:926

@vasireddy99 can you confirm?

majanjua-amzn: I will uncomment it regardless, had an offline discussion with Vasi and he said that it runs a test so might as well keep it in.

srprash (approved):

PR #905, opened 2024-10-09
feat: update bedrock guardrail cfn primary id

yiyuan-he:
### *Summary of Changes:*

- Updating the CFN Primary Id for Guardrail to be the ARN.  (Verified this in the AWS Cloudformation Console)

  Screenshot 2024-10-09 at 12 44 12 PM

- Updated `BEDROCKAGENTOPERATION` enum.
  - Our Bedrock Agent Runtime should support both Agent Operation and KnowledgeBased Operation [like in Java SDK v1](https://github.com/mxiamxia/opentelemetry-java-instrumentation/blob/965f7878c79c2be5851f23cf272edd39585f42af/instrumentation/aws-sdk/aws-sdk-1.11/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v1_11/AwsSdkExperimentalAttributesExtractor.java#L133).

Related changes in upstream package: https://github.com/mxiamxia/opentelemetry-java-instrumentation/pull/13

### *Testing Plan:*

Set up sample apps and manually verified correct span attributes were generated via auto-instrumentation. 

`Java SDK v1`
bedrock-guardrail-span-data-verification-v1

`Java SDK v2`
bedrock-guardrail-span-data-verification-v2

### DIFF
--- a/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
+++ b/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
@@ -44,6 +44,7 @@ import static software.amazon.opentelemetry.javaagent.providers.AwsAttributeKeys
 import static software.amazon.opentelemetry.javaagent.providers.AwsAttributeKeys.AWS_BUCKET_NAME;
 import static software.amazon.opentelemetry.javaagent.providers.AwsAttributeKeys.AWS_CLOUDFORMATION_PRIMARY_IDENTIFIER;
 import static software.amazon.opentelemetry.javaagent.providers.AwsAttributeKeys.AWS_DATA_SOURCE_ID;
+import static software.amazon.opentelemetry.javaagent.providers.AwsAttributeKeys.AWS_GUARDRAIL_ARN;
 import static software.amazon.opentelemetry.javaagent.providers.AwsAttributeKeys.AWS_GUARDRAIL_ID;
 import static software.amazon.opentelemetry.javaagent.providers.AwsAttributeKeys.AWS_KNOWLEDGE_BASE_ID;
 import static software.amazon.opentelemetry.javaagent.providers.AwsAttributeKeys.AWS_LAMBDA_NAME;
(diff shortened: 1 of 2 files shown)


awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -455,6 +456,8 @@ private static void setRemoteResourceTypeAndIdentifier(SpanData span, Attributes
         remoteResourceType = Optional.of(NORMALIZED_BEDROCK_SERVICE_NAME + "::Guardrail");
         remoteResourceIdentifier =
             Optional.ofNullable(escapeDelimiters(span.getAttributes().get(AWS_GUARDRAIL_ID)));
+        cloudformationPrimaryIdentifier =
+            Optional.ofNullable(escapeDelimiters(span.getAttributes().get(AWS_GUARDRAIL_ARN)));
```
zzhlogin, line 460: I think in upstream[ PR conversation](https://github.com/mxiamxia/opentelemetry-java-instrumentation/pull/13#discussion_r1795942737), we want to "extract remote_resource_identifier from cloudformation_primary_identifier to make sure the consistency of two values."? So the `AWS_GUARDRAIL_ID` won't be directly used here?

yiyuan-he: I think as long as the relationship between the `remote_resource_identifier` and `cloudformation_primary_identifier` is consistent with the pattern

```
remote_resource_identifier = <some_identifier>
cloudformation_primary_identifier = <arn_prefix>:<some_identifier>
```

it should be okay to directly assign these attributes instead of extracting if they are easily accessible from the request/response bodies. I think introducing logic to manually parse strings may increase risk of bugs in the future.

zzhlogin (dismissed):

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
mxiamxia, line 460: can we extract `remoteResourceIdentifier` from  `cloudformationPrimaryIdentifier` ARN to make sure the value consistency? I know most of time shouldn't be different but just in case

yiyuan-he: Okay sounds good. @zzhlogin was also suggesting this. I'll go with this approach and put up a new commit for it.

zzhlogin (approved):

mxiamxia (approved):

PR #944, opened 2024-11-04
Adding UdpExporter for Otlp spans

srprash:
### Description of changes:
Adding a new `SpanExporter` that can export spans to a configurable endpoint over UDP.
- The spans are first protobuf encoded, then base64 encoded, then prefixed with `T1S` or `T1U` (depending on sampled), and finally appended to the `{"format":"json","version":1}\n` string before being exported to the endpoint, default is `127.0.0.1:2000`.
- Related: https://github.com/aws-observability/aws-otel-python-instrumentation/pull/232

### Testing:
Wrote a test to create and export a real span via X-Ray Daemon running locally.
  - Unit test
    ```java
    @Test
    public void testExporter() { // TODO: only for testing. remove.
    Tracer tracer = OpenTelemetrySdk.builder().build().getTracer("hello");

    Span mySpan = tracer.spanBuilder("DoTheLoop_5").startSpan();
    int numIterations = 5;
    mySpan.setAttribute("NumIterations", numIterations);
    mySpan.setAttribute(AttributeKey.stringArrayKey("foo"), Arrays.asList("bar1", "bar2"));
    try (var scope = mySpan.makeCurrent()) {
      for (int i = 1; i <= numIterations; i++) {
        System.out.println("i = " + i);
      }
    } finally {
      mySpan.end();
    }

    OtlpUdpSpanExporter exporter = new OtlpUdpSpanExporterBuilder().build();

    ReadableSpan readableSpan = (ReadableSpan) mySpan;
    SpanData spanData = readableSpan.toSpanData();

    exporter.export(Collections.singletonList(spanData));
    }
    ```
  
  - Observed trace data in X-Ray service
     ```json
    {
    "Id": "1-246f2975-8c1d813535803fea548be818",
    "Duration": 0.003,
    "LimitExceeded": false,
    "Segments": [
        {
            "Id": "455654be8f47d669",
            "Document": {
                "id": "455654be8f47d669",
                "name": "DoTheLoop_5",
                "start_time": 1730828092.934588,
                "trace_id": "1-246f2975-8c1d813535803fea548be818",
                "end_time": 1730828092.9371202,
                "annotations": {
                    "span.name": "DoTheLoop_5",
                    "span.kind": "INTERNAL"
                },
                "metadata": {
                    "telemetry.sdk.language": "java",
                    "service.name": "unknown_service:java",
                    "foo": [
                        "bar1",
                        "bar2"
                    ],
                    "NumIterations": 5,
                    "telemetry.sdk.version": "1.34.1",
                    "telemetry.sdk.name": "opentelemetry"
                }
            }
        }
    ]
    }
    ```

### DIFF
--- a/awsagentprovider/build.gradle.kts
+++ b/awsagentprovider/build.gradle.kts
@@ -41,14 +41,14 @@ dependencies {
   // Export configuration
   compileOnly("io.opentelemetry:opentelemetry-exporter-otlp")
   // For Udp emitter
-  implementation("io.opentelemetry:opentelemetry-exporter-otlp-common")
-  implementation("io.opentelemetry:opentelemetry-exporter-common")
+  compileOnly("io.opentelemetry:opentelemetry-exporter-otlp-common")
 
   testImplementation("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure")
   testImplementation("io.opentelemetry:opentelemetry-sdk-testing")
   testImplementation("io.opentelemetry:opentelemetry-extension-aws")
   testImplementation("io.opentelemetry:opentelemetry-extension-trace-propagators")
   testImplementation("com.google.guava:guava")
+  testRuntimeOnly("io.opentelemetry:opentelemetry-exporter-otlp-common")
 
   compileOnly("com.google.code.findbugs:jsr305:3.0.2")
   testImplementation("org.mockito:mockito-core:5.3.1")
--- a/awsagentprovider/build.gradle.kts
+++ b/awsagentprovider/build.gradle.kts
@@ -40,6 +40,9 @@ dependencies {
   implementation("com.amazonaws:aws-java-sdk-core:1.12.773")
   // Export configuration
   compileOnly("io.opentelemetry:opentelemetry-exporter-otlp")
+  // For Udp emitter
+  implementation("io.opentelemetry:opentelemetry-exporter-otlp-common")
+  implementation("io.opentelemetry:opentelemetry-exporter-common")
 
   testImplementation("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure")
   testImplementation("io.opentelemetry:opentelemetry-sdk-testing")
--- a/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpUdpSpanExporter.java
+++ b/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpUdpSpanExporter.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.opentelemetry.javaagent.providers;
+
+import io.opentelemetry.exporter.internal.otlp.traces.TraceRequestMarshaler;
+import io.opentelemetry.sdk.common.CompletableResultCode;
+import io.opentelemetry.sdk.trace.data.SpanData;
+import io.opentelemetry.sdk.trace.export.SpanExporter;
+import java.io.ByteArrayOutputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+import java.util.Collection;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+public class OtlpUdpSpanExporter implements SpanExporter {
+
+  private static final Logger logger = Logger.getLogger(OtlpUdpSpanExporter.class.getName());
+  private static final String PROTOCOL_HEADER = "{\"format\": \"json\", \"version\": 1}";
+  private static final char PROTOCOL_DELIMITER = '\n';
+  private static final String FORMAT_OTEL_SAMPLED_TRACES_BINARY_PREFIX = "T1S";
+  private static final String FORMAT_OTEL_UNSAMPLED_TRACES_BINARY_PREFIX = "T1U";
+
+  private final AtomicBoolean isShutdown = new AtomicBoolean();
+
+  private final UdpSender sender;
+  private final boolean sampled;
+
+  OtlpUdpSpanExporter(UdpSender sender, boolean sampled) {
+    this.sender = sender;
+    this.sampled = sampled;
+  }
+
+  @Override
+  public CompletableResultCode export(Collection<SpanData> spans) {
+    if (isShutdown.get()) {
+      return CompletableResultCode.ofFailure();
+    }
+
+    TraceRequestMarshaler exportRequest = TraceRequestMarshaler.create(spans);
+    ByteArrayOutputStream baos = new ByteArrayOutputStream();
+    try {
+      exportRequest.writeBinaryTo(baos);
+      String payload =
+          PROTOCOL_HEADER
+              + PROTOCOL_DELIMITER
+              + (sampled
+                  ? FORMAT_OTEL_SAMPLED_TRACES_BINARY_PREFIX
+                  : FORMAT_OTEL_UNSAMPLED_TRACES_BINARY_PREFIX)
+              + Base64.getEncoder().encodeToString(baos.toByteArray());
+      sender.send(payload.getBytes(StandardCharsets.UTF_8));
+      System.out.println("Sending::: " + payload); // TODO: remove
+      return CompletableResultCode.ofSuccess();
+    } catch (Exception e) {
+      logger.log(Level.SEVERE, "Failed to export spans. Error: " + e.getMessage(), e);
+      return CompletableResultCode.ofFailure();
+    }
+  }
+
+  @Override
+  public CompletableResultCode flush() {
+    return CompletableResultCode.ofSuccess();
+  }
+
+  @Override
+  public CompletableResultCode shutdown() {
+    if (!isShutdown.compareAndSet(false, true)) {
+      logger.log(Level.INFO, "Calling shutdown() multiple times.");
+      return CompletableResultCode.ofSuccess();
+    }
+    return sender.shutdown();
+  }
+}
--- a/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpUdpSpanExporter.java
+++ b/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpUdpSpanExporter.java
@@ -28,12 +28,26 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.concurrent.Immutable;
 
+/**
+ * Exports spans via UDP, using OpenTelemetry's protobuf model. The protobuf modelled spans are
+ * Base64 encoded and prefixed with AWS X-Ray specific information before being sent over to {@link
+ * UdpSender}.
+ *
+ * <p>This exporter is NOT meant for generic use since the payload is prefixed with AWS X-Ray
+ * specific information.
+ */
 @Immutable
-public class OtlpUdpSpanExporter implements SpanExporter {
+class OtlpUdpSpanExporter implements SpanExporter {
 
   private static final Logger logger = Logger.getLogger(OtlpUdpSpanExporter.class.getName());
+
+  // The protocol header and delimiter is required for sending data to X-Ray Daemon or when running
+  // in Lambda.
+  // https://docs.aws.amazon.com/xray/latest/devguide/xray-api-sendingdata.html#xray-api-daemon
   private static final String PROTOCOL_HEADER = "{\"format\": \"json\", \"version\": 1}";
   private static final char PROTOCOL_DELIMITER = '\n';
+
+  // These prefixes help the backend identify if the spans payload is sampled or not.
   private static final String FORMAT_OTEL_SAMPLED_TRACES_BINARY_PREFIX = "T1S";
   private static final String FORMAT_OTEL_UNSAMPLED_TRACES_BINARY_PREFIX = "T1U";
 
--- a/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpUdpSpanExporter.java
+++ b/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpUdpSpanExporter.java
@@ -65,7 +65,6 @@ public class OtlpUdpSpanExporter implements SpanExporter {
                   : FORMAT_OTEL_UNSAMPLED_TRACES_BINARY_PREFIX)
               + Base64.getEncoder().encodeToString(baos.toByteArray());
       sender.send(payload.getBytes(StandardCharsets.UTF_8));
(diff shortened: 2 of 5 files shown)


awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/UdpSender.java
```diff
@@ -0,0 +1,66 @@
...
+import io.opentelemetry.sdk.common.CompletableResultCode;
+import java.io.IOException;
+import java.net.DatagramPacket;
+import java.net.DatagramSocket;
+import java.net.InetSocketAddress;
+import java.net.SocketException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class UdpSender {
```
bjrara, line 27: As for best practice, please add javadoc to describe the exporter.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/UdpSender.java
```diff
@@ -0,0 +1,66 @@
...
+  private final DatagramSocket socket;
+  private final InetSocketAddress endpoint;
+
+  public UdpSender(String host, int port) throws SocketException {
+    this.endpoint = new InetSocketAddress(host, port);
+    try {
+      socket = new DatagramSocket();
+    } catch (SocketException e) {
+      logger.log(Level.SEVERE, "Exception while instantiating UdpSender socket.", e);
+      throw e;
```
bjrara, line 39: Do we really want to make the entire auto-instrumentation to fail due to udp initialization failure? What if the customer has other exporters?

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpUdpSpanExporter.java
```diff
@@ -0,0 +1,99 @@
...
+  private static final Logger logger = Logger.getLogger(OtlpUdpSpanExporter.class.getName());
+  private static final String PROTOCOL_HEADER = "{\"format\": \"json\", \"version\": 1}";
+  private static final char PROTOCOL_DELIMITER = '\n';
+  private static final String FORMAT_OTEL_SAMPLED_TRACES_BINARY_PREFIX = "T1S";
+  private static final String FORMAT_OTEL_UNSAMPLED_TRACES_BINARY_PREFIX = "T1U";
+
+  private final AtomicBoolean isShutdown = new AtomicBoolean();
+
+  private final UdpSender sender;
+  private final boolean sampled;
```
bjrara, line 43: It's really confusing why sampled decision is at exporter level, not span level. Can you add comment on how it is used?

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/UdpExporterTest.java
```diff
@@ -0,0 +1,106 @@
...
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.trace.*;
+import io.opentelemetry.sdk.common.InstrumentationScopeInfo;
+import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.sdk.trace.data.SpanData;
+import io.opentelemetry.sdk.trace.data.StatusData;
+import java.util.Collections;
+import org.junit.jupiter.api.Test;
+
+public class UdpExporterTest {
```
bjrara, line 32: Can we add test cases to verify the content, e.g. protocol header, prefix...

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpUdpSpanExporter.java
```diff
@@ -0,0 +1,99 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+import java.util.Collection;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+public class OtlpUdpSpanExporter implements SpanExporter {
```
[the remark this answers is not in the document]
srprash: Sure. will do

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpUdpSpanExporter.java
```diff
@@ -0,0 +1,99 @@
...
+
+  private static final Logger logger = Logger.getLogger(OtlpUdpSpanExporter.class.getName());
+  private static final String PROTOCOL_HEADER = "{\"format\": \"json\", \"version\": 1}";
```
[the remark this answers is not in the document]
srprash: Since the exporter is primarily meant to export spans when on Lambda (which is similar to exporting to X-Ray Daemon), the Udp payload needs this header.
https://docs.aws.amazon.com/xray/latest/devguide/xray-api-sendingdata.html#xray-api-daemon

I will add a code comment for the same.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/UdpSender.java
srprash: I was on the fence for whether to fail or not. But you have a good point. There could be other exporters, so the auto-instrumentation should still work.
I will remove the thrown exception and only keep the log.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpUdpSpanExporter.java
srprash: I agree that it is confusing and unfortunately I don't have a great explanation for this. 
When on Lambda, the entire batch of spans will either be sampled or unsampled. Since Lambda execution is short lived, this sample information can be set when the OtlpUdpSpanExporter is being initialized once, used later during export.

For reference: https://github.com/aws-observability/aws-otel-python-instrumentation/pull/247

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/UdpExporterTest.java
srprash: Sure. Will do.

awsagentprovider/build.gradle.kts
```diff
@@ -40,6 +40,9 @@ dependencies {
   implementation("com.amazonaws:aws-java-sdk-core:1.12.773")
   // Export configuration
   compileOnly("io.opentelemetry:opentelemetry-exporter-otlp")
+  // For Udp emitter
+  implementation("io.opentelemetry:opentelemetry-exporter-otlp-common")
```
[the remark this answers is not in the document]
srprash: I'm not a gradle expert so I thought `implementation` (and thus availability at runtime) is needed for running unit tests.
But then I found using `compileOnly` and `testRuntimeOnly` for this is the right approach.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpUdpSpanExporter.java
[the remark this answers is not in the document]
srprash: Done. Also, made the class (and others as well) non-public since there is no need currently to exposed this functionality to users from outside of ADOT.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/UdpSender.java
```diff
@@ -0,0 +1,69 @@
...
+    try {
+      this.socket = new DatagramSocket();
+    } catch (SocketException e) {
+      logger.log(Level.SEVERE, "Exception while instantiating UdpSender socket.", e);
+    }
+  }
+
+  public CompletableResultCode shutdown() {
+    try {
+      socket.close();
```
bjrara, line 48: 1. `socket` can be null if it fails in constructor with SocketException. You might want to check it to avoid NPE.
2. Since we've caught `SocketException`, is it still needed to claim `throws SocketException` at constructor?

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpUdpSpanExporter.java
bjrara: Thanks for the reference. I see the decision is made on span in the Python version. The traffic is diverged at the leading span processor. Will we have the same solution in Java? Have we evaluated if the same solution can be applied?

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpUdpSpanExporter.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    TraceRequestMarshaler exportRequest = TraceRequestMarshaler.create(spans);
+    ByteArrayOutputStream baos = new ByteArrayOutputStream();
+    try {
+      exportRequest.writeBinaryTo(baos);
+      String payload =
+          PROTOCOL_HEADER
+              + PROTOCOL_DELIMITER
+              + (sampled
```
bjrara, line 77: Combining the discussion at [this thread](https://github.com/aws-observability/aws-otel-java-instrumentation/pull/944/files#r1829838239), my suggestion for you to evaluate would be

Option 1. Read the sampled decision from span if you want the exporter to work independently.
Option 2. Pass the predefined (un)sampled protocol prefix into exporter instead of a boolean `sampled`, and add comment on top with some description. I don't see why we should repeatedly build the same string in the exporting function.

srprash: I had a discussion with @wangzlei and he explained that the idea of using 2 processors is to segregate sampled and unsampled spans into respective batches for export. This ensures each payload contains either all sampled or unsampled spans. 

Given how the exporter receives a collection of spans in its `export()`, I think something like Option 2 (assuming the 2 BSP solution works in Java) would be better than having to separate spans in the exporter and send two payloads.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/UdpSender.java
srprash: Done. No we don't need the `throws SocketException` in the constructor signature.

wangzlei (approved):

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/UdpSender.java
```diff
@@ -0,0 +1,72 @@
...
+    } catch (Exception e) {
+      logger.log(Level.SEVERE, "Exception while closing UdpSender socket.", e);
+      return CompletableResultCode.ofFailure();
+    }
+  }
+
+  public void send(byte[] data) {
+    DatagramPacket packet = new DatagramPacket(data, data.length, endpoint);
+    try {
+      socket.send(packet);
```
bjrara, line 62: Shall we check null in `send`?

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpUdpSpanExporter.java
srprash: @bjrara - BSP has [this flag](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorBuilder.java#L45) which enables exporting of unsampled spans too.
We can have a custom SpanProcessor delegate the unsampled spans to a BSP with `exportUnsampledSpans` enabled.

srprash: Did a quick and dirty test to validate the above works.

#### Setup
##### Custom SpanProcessor `AwsUnsampledSpanProcessor`:
```java
package software.amazon.opentelemetry.javaagent.providers;

import io.opentelemetry.context.Context;
import io.opentelemetry.sdk.common.CompletableResultCode;
import io.opentelemetry.sdk.trace.ReadWriteSpan;
import io.opentelemetry.sdk.trace.ReadableSpan;
import io.opentelemetry.sdk.trace.SpanProcessor;
import io.opentelemetry.sdk.trace.export.BatchSpanProcessor;
import io.opentelemetry.sdk.trace.export.SpanExporter;

public class AwsUnsampledSpanProcessor implements SpanProcessor {
    public final SpanProcessor forwardingBSP;

    AwsUnsampledSpanProcessor(SpanExporter exporter) {
        this.forwardingBSP = BatchSpanProcessor.builder(exporter)
                .setExportUnsampledSpans(true)
                .build();
    }

    @Override
    public void onStart(Context context, ReadWriteSpan readWriteSpan) {
        forwardingBSP.onStart(context, readWriteSpan);
    }

    @Override
    public boolean isStartRequired() {
        return forwardingBSP.isStartRequired();
    }

    @Override
    public void onEnd(ReadableSpan readableSpan) {
        if (readableSpan.getSpanContext().isSampled()) {
            return;
        }
        forwardingBSP.onEnd(readableSpan);
    }

    @Override
    public boolean isEndRequired() {
        return forwardingBSP.isEndRequired();
    }

    @Override
    public CompletableResultCode shutdown() {
        return forwardingBSP.shutdown();
    }

    @Override
    public CompletableResultCode forceFlush() {
        return forwardingBSP.forceFlush();
    }

    @Override
    public void close() {
        forwardingBSP.close();
    }
}
```
##### Code to create a unsampled span and process via the `AwsUnsampledSpanProcessor`
```java
@Test
  public void testExportingUnsampledSpansWithCustomProcessor() {

    SpanExporter unsampledSpanExporter = new OtlpUdpSpanExporterBuilder()
            .setPayloadSampleDecision(TracePayloadSampleDecision.UNSAMPLED)
            .build();
    SpanProcessor unsampledSpanProcessor = new AwsUnsampledSpanProcessor(unsampledSpanExporter);

    SdkTracerProvider tp = SdkTracerProvider.builder()
            .addSpanProcessor(unsampledSpanProcessor)
            .setSampler(AlwaysRecordSampler.create(Sampler.alwaysOff())) //set sampler
            .build();
    Tracer tracer = OpenTelemetrySdk.builder().setTracerProvider(tp).build().getTracer("hello");

    Span mySpan = tracer.spanBuilder("DoTheLoop_8").startSpan();
    int numIterations = 5;
    mySpan.setAttribute("NumIterations", numIterations);
    mySpan.setAttribute(AttributeKey.stringArrayKey("foo"), Arrays.asList("bar1", "bar2"));
    try (var scope = mySpan.makeCurrent()) {
      for (int i = 1; i <= numIterations; i++) {
        System.out.println("i = " + i);
      }
    } finally {
      mySpan.end();
    }

    unsampledSpanProcessor.forceFlush();
  }
```

#### Observations
- When spans are unsampled the payload is exported with `T1U` prefix
- When changing the sampler to `alwaysOn`, no spans are exported. This is expected since the only SpanProcessor added in my setup is the `AwsUnsampledSpanProcessor` so the sampled spans are simply dropped.
- This shows that the `AwsUnsampledSpanProcessor` can be used to separate and export only the unsampled spans.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/UdpSender.java
srprash: done

bjrara (approved):

srprash:
The trivy image scan failure causing CI to fail is due to this issue: https://github.com/aquasecurity/trivy-action/issues/389
No fix from trivy yet. A suggested workaround is to pull the db image from AWS ECR but I haven't tried it.

srprash:
Had to override the checks and merge due to trivy scan issue mentioned above.

PR #948, opened 2024-11-07
Adding AwsUnsampledOnlySpanProcessor to export batches of unsampled spans

srprash:
#### Description of changes

Following the approach as ADOT Python: https://github.com/aws-observability/aws-otel-python-instrumentation/pull/247. Note that the span attribute key was [later changed to `aws.trace.flag.sampled`](https://github.com/aws-observability/aws-otel-python-instrumentation/pull/260).

Introducing `AwsUnsampledOnlySpanProcessor` that can be appended to a `TracerProvider`. This span processor wraps a `BatchSpanProcessor` for delegating all the operations after processing. The `AwsUnsampledOnlySpanProcessor` does 2 things:
  - onStart - Sets the `aws.trace.flag.sampled` span attribute on each span.
  - onEnd - If the span is unsampled, only then it is delegated to the BSP, otherwise it is dropped.

##### Setup
```
@Test
  public void testExportingUnsampledSpan() {
    SpanProcessor unsampledSpanProcessor = AwsUnsampledOnlySpanProcessor.builder().build();

    SdkTracerProvider tp =
        SdkTracerProvider.builder()
            .addSpanProcessor(unsampledSpanProcessor)
            .setSampler(AlwaysRecordSampler.create(Sampler.alwaysOff())) // change to alwaysOn to sample
            .build();

    Tracer tracer = OpenTelemetrySdk.builder().setTracerProvider(tp).build().getTracer("hello");

    Span span = tracer.spanBuilder("MySpan_unsampled_006").startSpan();
    span.setAttribute("foo", "bar");
    span.end();

    unsampledSpanProcessor.forceFlush();

    // add 5 seconds delay
    try {
      Thread.sleep(5000);
    } catch (InterruptedException e) {
      e.printStackTrace();
    }
  }
```

##### Observations:
- Unsampled span is exported in a batch.
- Switching to `alwaysOn` sampler, the sampled span is not exported at all.

### DIFF
--- a/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsUnsampledOnlySpanProcessor.java
+++ b/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsUnsampledOnlySpanProcessor.java
@@ -57,12 +57,12 @@ final class AwsUnsampledOnlySpanProcessor implements SpanProcessor {
 
   @Override
   public boolean isStartRequired() {
-    return delegate.isStartRequired();
+    return true;
   }
 
   @Override
   public boolean isEndRequired() {
-    return delegate.isEndRequired();
+    return true;
   }
 
   @Override
--- a/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsUnsampledOnlySpanProcessor.java
+++ b/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsUnsampledOnlySpanProcessor.java
@@ -21,6 +21,10 @@ import io.opentelemetry.sdk.trace.ReadWriteSpan;
 import io.opentelemetry.sdk.trace.ReadableSpan;
 import io.opentelemetry.sdk.trace.SpanProcessor;
 
+/**
+ * {@link SpanProcessor} that only exports unsampled spans in a batch via a delegated @{link BatchSpanProcessor}.
+ * The processor also adds an attribute to each processed span to indicate that it was sampled or not.
+ */
 final class AwsUnsampledOnlySpanProcessor implements SpanProcessor {
 
   private final SpanProcessor delegate;
--- a/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsUnsampledOnlySpanProcessor.java
+++ b/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsUnsampledOnlySpanProcessor.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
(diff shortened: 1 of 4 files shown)


awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsUnsampledOnlySpanProcessor.java
```diff
@@ -0,0 +1,87 @@
...
+  public static AwsUnsampledOnlySpanProcessorBuilder builder() {
+    return new AwsUnsampledOnlySpanProcessorBuilder();
+  }
+
+  @Override
+  public void onStart(Context parentContext, ReadWriteSpan span) {
+    if (span.getSpanContext().isSampled()) {
+      span.setAttribute(AwsAttributeKeys.AWS_TRACE_FLAG_SAMPLED, true);
+    } else {
+      span.setAttribute(AwsAttributeKeys.AWS_TRACE_FLAG_SAMPLED, false);
```
wangzlei, line 46: We don't need FLAG_SAMPLED=false if span is sampled, it can save span size a little bit

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsUnsampledOnlySpanProcessor.java
```diff
@@ -0,0 +1,87 @@
...
+import io.opentelemetry.sdk.trace.SpanProcessor;
+
+/**
+ * {@link SpanProcessor} that only exports unsampled spans in a batch via a delegated @{link
+ * BatchSpanProcessor}. The processor also adds an attribute to each processed span to indicate that
+ * it was sampled or not.
+ */
+final class AwsUnsampledOnlySpanProcessor implements SpanProcessor {
+
+  private final SpanProcessor delegate;
```
wangzlei, line 31: The delegate would be BatchSpanProcessor with exportUnsampledSpans=true?
https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java#L56

srprash: Yes. The default build of `AwsUnsampledOnlySpanProcessor` has this as its delegate:
```
public AwsUnsampledOnlySpanProcessor build() {
    BatchSpanProcessor bsp =
        BatchSpanProcessor.builder(exporter).setExportUnsampledSpans(true).build();
    return new AwsUnsampledOnlySpanProcessor(bsp);
  }
```

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsUnsampledOnlySpanProcessor.java
srprash: I see your point and I can make the change to align with current implementations in Python and JavaScript.

However, as a boolean span attribute, `aws.trace.flag.sampled` should always be present for consistency. Without `aws.trace.flag.sampled=false`, it could be unclear whether the span was unsampled or if there was a malfunction in the SpanProcessor and the span was actually sampled. Ensuring this attribute is always present would improve clarity. Let me know if you agree, and I can assist with implementing this change in Python and JavaScript.

wangzlei: It is ok to have no this attribute since by default the span in backend should be sampled.
Having this attribute actually is specific to a CloudWatch feature, it means backend can add true if want to keep consistency for user experience.

srprash: ok. made the change.

wangzlei (approved):

PR #953, opened 2024-11-18
Utility for building a lambda layer artifact

srprash:
## Description of changes
- Adding a script and utilities to build a `.zip` artifact that will be published as the ADOT Java Lambda Layer.
- The layer packages patched version of upstream (opentelemetry-java-instrumentation) and the current source build of aws-opentelemetry-java-instrumentation, including the Application Signals logic.
- The layer building logic is primarily borrowed from what we have in the [aws-otel-lambda](https://github.com/aws-observability/aws-otel-lambda/tree/main/java) repo, and is tweaked to the build a javaagent instrumentation only layer from the current repo.

### Setup
- run `lambda-layer/build-layer.sh`
  - This will produce a `opentelemetry-javaagent-layer.zip` artifact.
- From the AWS Lambda console, created a Java 17 function. The code makes a remote call using OkHttp to `www.amazon.com`
- From the AWS Lambda console, create a lambda layer using the `opentelemetry-javaagent-layer.zip` file obtained in previous step.
- Set the following environment variables on the lambda function:
  ```
  AWS_LAMBDA_EXEC_WRAPPER=/opt/otel-handler
  OTEL_AWS_APPLICATION_SIGNALS_ENABLED=true
  OTEL_LOGS_EXPORTER=none
  OTEL_METRICS_EXPORTER=none
  OTEL_TRACES_EXPORTER=otlp
  ```
- Enable active tracing on the function
- Invoke the function from lambda console

#### Trace view
image

#### Trace timeline view
image

#### Trace raw data
```
{
    "Id": "1-673ccffc-3f2d31933a2b0f341ee61980",
    "Duration": 7.728,
    "LimitExceeded": false,
    "Segments": [
        {
            "Id": "8aba0f96e08d63a1",
            "Document": {
                "id": "8aba0f96e08d63a1",
                "name": "java-lambda-example",
                "start_time": 1732038657.532657,
                "trace_id": "1-673ccffc-3f2d31933a2b0f341ee61980",
                "end_time": 1732038660.474236,
                "parent_id": "31aa4a42384b852d",
                "aws": {
                    "cloudwatch_logs": [
                        {
                            "log_group": "/aws/lambda/java-lambda-example"
                        }
                    ]
                },
                "annotations": {
                    "aws:responseLatency": 2939.22,
                    "aws:runtimeOverhead": 1.565,
                    "aws:responseDuration": 0.101
                },
                "origin": "AWS::Lambda::Function",
                "subsegments": [
                    {
[...]

PR #956, opened 2024-11-25
Renaming to otel-instrument and default configs

srprash:
As a follow up to the https://github.com/aws-observability/aws-otel-java-instrumentation/pull/953 and from internal discussion with @wangzlei:
- Renaming `otel-handler` to `otel-instrument` to be consistent with other ADOT lambda layer script naming.
- Enabling Application Signals by default with the following environment variables already set. Note that users can still override these environment variables if they want to.
  ```
  AWS_LAMBDA_EXEC_WRAPPER=/opt/otel-handler
  OTEL_AWS_APPLICATION_SIGNALS_ENABLED=true
  OTEL_LOGS_EXPORTER=none
  OTEL_METRICS_EXPORTER=none
  OTEL_TRACES_EXPORTER=otlp
  ```
- Appending Lambda resource attributes

ADOT Python implementation for reference: https://github.com/aws-observability/aws-otel-python-instrumentation/blame/main/lambda-layer/src/otel-instrument

### DIFF
--- a/lambda-layer/build-layer.sh
+++ b/lambda-layer/build-layer.sh
@@ -45,4 +45,4 @@ popd
 
 ## Copy ADOT Java Agent downloaded using Gradle task and bundle it with the Lambda handler script
 cp "$SOURCEDIR"/build/javaagent/aws-opentelemetry-agent*.jar ./opentelemetry-javaagent.jar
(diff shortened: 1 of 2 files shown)


wangzlei (approved):
LGTM!

PR #982, opened 2024-12-23
Use grpc exporter for agent test workflow.

zzhlogin:
*Description of changes:*
Add `OTEL_EXPORTER_OTLP_PROTOCOL` to `grpc` for instrumentation as in [version v2.0.0](https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v2.0.0):

> The default OTLP protocol has been changed from grpc to http/protobuf in orderto align with the specification.

### DIFF
--- a/.github/collector/docker-compose.yml
+++ b/.github/collector/docker-compose.yml
@@ -26,6 +26,7 @@ services:
       - AWS_REGION=us-west-2
       - OTEL_JAVAAGENT_DEBUG=true
       - OTEL_METRICS_EXPORTER=otlp
+      - OTEL_EXPORTER_OTLP_PROTOCOL=grpc
     volumes:
       - /tmp/awscreds:/tmp/awscreds
     ports:


srprash:
Can you explain why do we need to explicitly export using grpc and not the default http?

zzhlogin:
> Can you explain why do we need to explicitly export using grpc and not the default http?

Previously, in v1.x, the default protocol is grpc, and the smoke tests is using the default protocol [here](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/1889c91d4868fd192108b0d2807744074048b744/.github/collector/collector-config.yml#L4). 

Starting from upstream v2.x, the default OTLP protocol has been changed from grpc to http/protobuf ([release note](https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v2.0.0)). So that we need to specifically set the protocol using env `OTEL_EXPORTER_OTLP_PROTOCOL` to align with smoke test setup.

srprash:
Thanks. It makes sense now. :)

srprash (approved):

PR #986, opened 2024-12-26
Updating Lambda Patch v1.33.6 + Fast Startup capability

srprash:
### Description of changes:
- Fast Startup feature was added as opt-in in the upstream some time ago: https://github.com/open-telemetry/opentelemetry-lambda/pull/1180 (includes perf improvement results)
  - Did some of my own testing on a Java17 Lambda function with 3GB of memory and noticed cold start time improvement from 8.7 seconds to 4.2 seconds.

- Since the ADOT 1.x version has been updated to lated 1.33.6 version of upstream, we need to update the lambda layer related patches to this version as well. 
  - Tested by manually building the layer and instrumenting a sample lambda function. 

### DIFF
--- a/lambda-layer/patches/opentelemetry-java-instrumentation.patch
+++ b/lambda-layer/patches/opentelemetry-java-instrumentation.patch
@@ -636,10 +636,10 @@ index cc1414c0bf..db8a59b046 100644
 --- a/version.gradle.kts
 +++ b/version.gradle.kts
 @@ -1,5 +1,5 @@
--val stableVersion = "1.32.1"
--val alphaVersion = "1.32.1-alpha"
-+val stableVersion = "1.32.1-adot-lambda1"
-+val alphaVersion = "1.32.1-adot-lambda1-alpha"
+-val stableVersion = "1.33.6"
+-val alphaVersion = "1.33.6-alpha"
++val stableVersion = "1.33.6-adot-lambda1"
++val alphaVersion = "1.33.6-adot-lambda1-alpha"
 
  allprojects {
    if (findProperty("otel.stable") != "true") {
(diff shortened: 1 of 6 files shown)


lambda-layer/patches/opentelemetry-java-instrumentation.patch
```diff
@@ -636,10 +636,10 @@ index cc1414c0bf..db8a59b046 100644
 --- a/version.gradle.kts
 +++ b/version.gradle.kts
 @@ -1,5 +1,5 @@
--val stableVersion = "1.32.1"
--val alphaVersion = "1.32.1-alpha"
-+val stableVersion = "1.32.1-adot-lambda1"
-+val alphaVersion = "1.32.1-adot-lambda1-alpha"
+-val stableVersion = "1.33.6"
+-val alphaVersion = "1.33.6-alpha"
++val stableVersion = "1.33.6-adot-lambda1"
```
zzhlogin, line 641: In current build-layer.sh, the [lambda layer](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/dc6ba3742a81f6dce2b15965a1012f42cfd9a83e/lambda-layer/build-layer.sh#L23) is applied first, then the apply the adot patch. Whereas in [adot patch](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/dc6ba3742a81f6dce2b15965a1012f42cfd9a83e/.github/patches/opentelemetry-java-instrumentation.patch#L1360) the version number is changed from:
```
-val stableVersion = "1.33.6"
-val alphaVersion = "1.33.6-alpha"
```
to 
```
+val stableVersion = "1.33.6-adot1"
+val alphaVersion = "1.33.6-adot1-alpha"
```
Does this mean the lambda patch should continue use the following version numbers?
```
-val stableVersion = "1.33.6"
-val alphaVersion = "1.33.6-alpha"
```

srprash: Good callout. As we discussed offline, I have reversed the patching order and slightly modified the lambda layer patch to not conflict (and get rejected) anymore. Much cleaner this way. :)

zzhlogin (approved):

bjrara (approved):

PR #992, opened 2025-01-10
decouple apigw-lambda sample app from adot project build

srprash:
The sample app doesn't need ADOT Javaagent instrumentation, rather it depends on instrumentaiton via lambda layer. SO it makes sense to not depend on adot javaagent build for building this sample app.


mxiamxia (approved):
make sense! thanks

PR #1000, opened 2025-01-14
[Bug fixes] Lambda - duplicate lambda spans + appsignals from unsampled spans

srprash:
### Issue 1 - Duplicate lambda root spans
The instrumentation produces 2 lambda root spans upon each invocation of the function. This is a known issue in OTel: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/7808

#### Fix
- Otel has fixed it in [this PR](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/10736) but the change wasn't ported to the 1.x versions. So I took the diff from the upstream commit and made it into a patch that we apply when building the ADOT Java Lambda Layer for v1.33.x.

### Issue 2 - Unsampled spans do not produce Application Signals metrics
On lambda environment, we export 100% of the spans to X-Ray to ensure we are able to provide 100% Application Signals metrics. However, currently only the sampled spans show up on the "Services" page and the unsampled spans do not.

#### Fix
- Upon comparing the sampled vs unsampled spans, I noticed that the unsampled spans are missing the attributes like `aws.local.service` and `aws.local.operation` which are required to generate Application Signals metrics.
- The fix is to wrap the `OtlpUdpSpanExporter` instance for unsampled spans with the `AwsMetricAttributesSpanExporter` so that the exported spans have the desired attributes.

#### Testing
- After creating a layer with the fix, I set the `OTEL_TRACES_SAMPLER` to `always_off`. Then I invoked the function once.
- The metrics appeared on the Application Signals console.
- See the following screenshots
  
image

image

*Description of changes:*

### DIFF
--- a/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsApplicationSignalsCustomizerProvider.java
+++ b/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsApplicationSignalsCustomizerProvider.java
@@ -206,7 +206,8 @@ public class AwsApplicationSignalsCustomizerProvider
 
       // If running on Lambda, we just need to export 100% spans and skip generating any Application
       // Signals metrics.
-      if (isLambdaEnvironment()) {
+      if (isLambdaEnvironment()
+          && System.getenv(OTEL_EXPORTER_OTLP_TRACES_ENDPOINT_CONFIG) == null) {
         String tracesEndpoint =
             Optional.ofNullable(System.getenv(AWS_XRAY_DAEMON_ADDRESS_CONFIG))
                 .orElse(DEFAULT_UDP_ENDPOINT);
--- a/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsApplicationSignalsCustomizerProvider.java
+++ b/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsApplicationSignalsCustomizerProvider.java
@@ -207,8 +207,25 @@ public class AwsApplicationSignalsCustomizerProvider
       // If running on Lambda, we just need to export 100% spans and skip generating any Application
       // Signals metrics.
       if (isLambdaEnvironment()) {
+        String tracesEndpoint =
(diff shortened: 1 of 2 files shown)


awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsApplicationSignalsCustomizerProvider.java
```diff
@@ -207,8 +207,25 @@ private SdkTracerProviderBuilder customizeTracerProviderBuilder(
...
+        String tracesEndpoint =
+            Optional.ofNullable(System.getenv(AWS_XRAY_DAEMON_ADDRESS_CONFIG))
+                .orElse(DEFAULT_UDP_ENDPOINT);
+        SpanExporter spanExporter =
+            new OtlpUdpSpanExporterBuilder()
+                .setPayloadSampleDecision(TracePayloadSampleDecision.UNSAMPLED)
+                .setEndpoint(tracesEndpoint)
+                .build();
+
+        // Wrap the udp exporter with the AwsMetricsAttributesSpanExporter to add Application
```
AsakerMohd, line 219: This is important. Good catch.

mxiamxia: In Python and NodeJS, if customer has set the customized OTLP endpoint through Env Var, we will not enable UDP exporter to send unsampled Span to the custom endpoint. Should we do the same for Java and .Net

JS: https://github.com/mxiamxia/aws-otel-js-instrumentation/blob/1bf42f613a3bc2142e396c5fabaef4f5245e0495/aws-distro-opentelemetry-node-autoinstrumentation/src/aws-opentelemetry-configurator.ts#L441C10-L441C36

Python: https://github.com/aws-observability/aws-otel-python-instrumentation/blob/main/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/aws_opentelemetry_configurator.py#L314

srprash: Yes. good callout. I will add that condition.

mxiamxia (dismissed):

mxiamxia (approved):

PR #1024, opened 2025-02-24
Fixed toString for OtlpAwsSpanExporter

liustve:
*Description of changes:*
Modified the toString method for the SigV4 exporter to say OtlpAwsSpanExporter instead of OtlpHttpSpanExporter.

### DIFF
--- a/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpAwsSpanExporter.java
+++ b/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpAwsSpanExporter.java
@@ -103,7 +103,7 @@ public class OtlpAwsSpanExporter implements SpanExporter {
 
   @Override
   public String toString() {
-    return this.parentExporter.toString();
+    return this.parentExporter.toString().replace("OtlpHttpSpanExporter", "OtlpAwsSpanExporter");
   }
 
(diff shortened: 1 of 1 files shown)


awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpAwsSpanExporter.java
```diff
@@ -103,7 +103,7 @@ public CompletableResultCode shutdown() {
 
   @Override
   public String toString() {
-    return this.parentExporter.toString();
+    return this.parentExporter.toString().replace("OtlpHttpSpanExporter", "OtlpAwsSpanExporter");
```
bjrara, line 106: Why not return a constant value?

liustve: I wanted to keep the same toString message as this class:

https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/otlp/all/src/main/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporter.java

But override the class name to make it more clear for log messages.

liustve: toString copied in new PR:

https://github.com/aws-observability/aws-otel-java-instrumentation/pull/1046

PR #1046, opened 2025-03-12
SigV4 Authentication support for http exporter

liustve:
**Background**
Supporting ADOT auto instrumentation to automatically inject SigV4 authentication headers for outgoing export trace requests to the allow exporting to the AWS XRay OTLP endpoint. Users will need to configure the following environment variables in order to enable and properly run this exporter:

OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=https://xray.[AWS-REGION].amazonaws.com/v1/traces; **required**
OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=http/protobuf **required**
OTEL_LOGS_EXPORTER=none; 
OTEL_METRICS_EXPORTER=none;

*Description of changes:*
1. Added new OtlpAwsSpanExporter class which uses composition to extend upstream's OtlpHttpSpanExporter which is responsible for making the http/grpc client calls to export traces to the given endpoint.

2. The OtlpAwsSpanExporter customizes the headers by adding an intermediary step to sign the request with SigV4 authentication and injects the signed headers to the outgoing trace request

3. In order to ensure we don't override any user configurations from environment variables, the OtlpAwsSpanExporter constructor copies any existing configurations create by upstream's auto instrumentation except for the following environment variable: OTEL_EXPORTER_OTLP_TRACES_HEADERS

Any headers set here will not be acknowledged within this exporter.

4. The ADOT auto instrumentation is now configured to automatically detect if a user is exporting to XRay OTLP endpoint by checking if the environment variable OTEL_EXPORTER_OTLP_TRACES_ENDPOINT is configured to match this url pattern:
https://xray.[AWS-REGION].amazonaws.com/v1/traces

AND by checking if 
OTEL_EXPORTER_OTLP_TRACES_PROTOCOL is set to http/protobuf

**Testing:**
1. Unit tests were added to ensure that the behavior of the exporter was properly injecting SigV4 headers before making the outgoing request
2. Manual testing was done by configuring the above environment variables and setting up the sample app locally with ADOT auto instrumentation and verified the spans in CW Logs.
3. The sample app was run and rerun 30 times and confirmed issues exporting the traces to the endpoint

Further testing will be done with the Release tests.


awsagentprovider/build.gradle.kts
```diff
@@ -45,12 +45,17 @@ dependencies {
   // For Udp emitter
   compileOnly("io.opentelemetry:opentelemetry-exporter-otlp-common")
 
+  // For OtlpAwsSpanExporter SigV4 Authentication
+  implementation("software.amazon.awssdk:auth")
+  implementation("software.amazon.awssdk:http-auth-aws")
```
majanjua-amzn, line 50: What's the difference between `compileOnly` and `implementation` here? Is `implementation` required and better?

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsApplicationSignalsCustomizerProvider.java
```diff
@@ -95,6 +98,9 @@ public class AwsApplicationSignalsCustomizerProvider
   private static final String OTEL_JMX_TARGET_SYSTEM_CONFIG = "otel.jmx.target.system";
   private static final String OTEL_EXPORTER_OTLP_TRACES_ENDPOINT_CONFIG =
       "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT";
+  private static final String OTEL_EXPORTER_HTTP_PROTOBUF_PROTOCOL = "http/protobuf";
+  private static final String OTEL_EXPORTER_OTLP_TRACES_PROTOCOL_CONFIG =
```
majanjua-amzn, line 102: This is new right? Why do we need it?

I don't think we should use env variables that look like they're configs from upstream OTEL, i.e. `OTEL_...`, unless they actually exist in upstream.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsApplicationSignalsCustomizerProvider.java
```diff
@@ -121,6 +129,37 @@ static boolean isLambdaEnvironment() {
...
+      logger.log(Level.INFO, "Detected using AWS OTLP XRay Endpoint.");
+
+      String otlpTracesProtocol = System.getenv(OTEL_EXPORTER_OTLP_TRACES_PROTOCOL_CONFIG);
+
+      if (otlpTracesProtocol == null
+          || !otlpTracesProtocol.equals(OTEL_EXPORTER_HTTP_PROTOBUF_PROTOCOL)) {
+        logger.info(
+            String.format(
+                "Improper configuration: Please configure your environment variables and export/set %s=%s",
+                OTEL_EXPORTER_OTLP_TRACES_PROTOCOL_CONFIG, OTEL_EXPORTER_HTTP_PROTOBUF_PROTOCOL));
```
majanjua-amzn, line 150: nit:
```suggestion
        logger.error(
            String.format(
                "Improper configuration: Please configure your environment variables and export/set %s=%s",
                OTEL_EXPORTER_OTLP_TRACES_PROTOCOL_CONFIG, OTEL_EXPORTER_HTTP_PROTOBUF_PROTOCOL));
```

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpAwsSpanExporter.java
```diff
@@ -0,0 +1,163 @@
...
+ * to the XRay OTLP endpoint https://xray.[AWSRegion].amazonaws.com/v1/traces. Utilizes the AWSSDK
+ * library to sign and directly inject SigV4 Authentication to the exported request's headers. <a
+ * href="https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-OTLPEndpoint.html">...</a>
+ */
+@Immutable
+public class OtlpAwsSpanExporter implements SpanExporter {
+  private static final String SERVICE_NAME = "xray";
+  private static final Logger logger = LoggerFactory.getLogger(OtlpAwsSpanExporter.class);
+
+  private final OtlpHttpSpanExporterBuilder parentExporterBuilder;
```
majanjua-amzn, line 52: This is only needed for `toString()` right? Seems unnecessary otherwise

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpAwsSpanExporter.java
```diff
@@ -0,0 +1,163 @@
...
+                .setMemoryMode(MemoryMode.IMMUTABLE_DATA)
+                .setEndpoint(endpoint)
+                .setHeaders(new SigV4AuthHeaderSupplier());
+
+    this.parentExporter = this.parentExporterBuilder.build();
+
+    this.awsRegion = endpoint.split("\\.")[1];
+    this.endpoint = endpoint;
+    this.spanData = new ArrayList<>();
+  }
```
majanjua-amzn, line 82: Shouldn't create the regular builder unless it's needed. Feel free to use `else` instead of `return` to ensure only one of those paths runs.
```suggestion
  public OtlpAwsSpanExporter(OtlpHttpSpanExporter parentExporter, String endpoint) {
    this.awsRegion = endpoint.split("\\.")[1];
    this.endpoint = endpoint;
    this.spanData = new ArrayList<>();
    
    if (parentExporter == null) {
        this.parentExporter =
            OtlpHttpSpanExporter.builder()
                .setMemoryMode(MemoryMode.IMMUTABLE_DATA)
                .setEndpoint(endpoint)
                .setHeaders(new SigV4AuthHeaderSupplier());
        return;
    }
    this.parentExporterBuilder =
        parentExporter.toBuilder()
            .setMemoryMode(MemoryMode.IMMUTABLE_DATA)
            .setEndpoint(endpoint)
            .setHeaders(new SigV4AuthHeaderSupplier());
  }
```

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpAwsSpanExporter.java
```diff
@@ -0,0 +1,163 @@
...
+  private Collection<SpanData> spanData;
+
+  public OtlpAwsSpanExporter(String endpoint) {
+    this(null, endpoint);
+  }
+
+  public OtlpAwsSpanExporter(OtlpHttpSpanExporter parentExporter, String endpoint) {
+    OtlpHttpSpanExporterBuilder defaultExporterBuilder =
+        OtlpHttpSpanExporter.builder()
+            .setMemoryMode(MemoryMode.IMMUTABLE_DATA)
```
majanjua-amzn, line 65: Looking forward to the investigation of the implications of this change.

awsagentprovider/build.gradle.kts
liustve: compileOnly will allow you to use these libraries without error when you write the code but will not load them during runtime. I need implementation here to load these libraries at runtime during the auto instrumentation startup.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsApplicationSignalsCustomizerProvider.java
liustve: The OTEL_EXPORTER_OTLP_TRACES_PROTOCOL_CONFIG is a config from upstream:
https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_protocol

 I got rid of the "instanceof OtlpHttpSpanExporter" check so I need to first verify that they're using the correct protocol and the correct exporter.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsApplicationSignalsCustomizerProvider.java
```diff
@@ -221,6 +260,10 @@ private SdkTracerProviderBuilder customizeTracerProviderBuilder(
         return tracerProviderBuilder;
       }
 
+      if (isSigV4Enabled) {
```
bjrara, line 263: Not following. Why we pass all the remaining initialization flow if SigV4 is enabled.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsApplicationSignalsCustomizerProvider.java
```diff
@@ -121,6 +129,37 @@ static boolean isLambdaEnvironment() {
...
+    boolean isXrayOtlpEndpoint =
+        otlpEndpoint != null
+            && Pattern.compile(XRAY_OTLP_ENDPOINT_PATTERN)
+                .matcher(otlpEndpoint.toLowerCase())
+                .matches();
+
+    if (isXrayOtlpEndpoint) {
+      logger.log(Level.INFO, "Detected using AWS OTLP XRay Endpoint.");
+
+      String otlpTracesProtocol = System.getenv(OTEL_EXPORTER_OTLP_TRACES_PROTOCOL_CONFIG);
```
bjrara, line 143: Should this validation happen earlier before the regular expression matches? If it's not `http/protobuf`, we won't need to parse the endpoint and check.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpAwsSpanExporter.java
```diff
@@ -0,0 +1,163 @@
...
+  private final String awsRegion;
+  private final String endpoint;
+  private Collection<SpanData> spanData;
+
+  public OtlpAwsSpanExporter(String endpoint) {
+    this(null, endpoint);
+  }
+
+  public OtlpAwsSpanExporter(OtlpHttpSpanExporter parentExporter, String endpoint) {
+    OtlpHttpSpanExporterBuilder defaultExporterBuilder =
```
bjrara, line 63: Why do we need to create `defaultExporterBuilder` if `parentExporter` is not null?

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/OtlpAwsSpanExporter.java
```diff
@@ -0,0 +1,170 @@
...
+        return result;
+
+      } catch (Exception e) {
+        logger.log(
+            Level.SEVERE,
+            String.format(
+                "Failed to sign/authenticate the given exported Span request to OTLP CloudWatch endpoint with error: %s",
+                e.getMessage()));
+
+        return new HashMap<>();
```
srprash, line 166: This is more efficient than initializing a new object.
```suggestion
        return Collections.emptyMap();
```

awsagentprovider/build.gradle.kts
srprash: +1 to Mengyi's comment

srprash (approved):
LGTM. Please address the remaining comments before merging.

PR #1047, opened 2025-03-13
Port upstream fix (PR#10736) to downstream v1 Lambda stream handler

lukeina2z:
- **V2 Issue:** [Non-top-level server span created by Lambda instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/7808)  

- **V2 PR:** [#10736](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/10736)  

The fix has been ported to v1 as a patch for non-stream handlers. This change extends the fix to the stream handler in the v1 repository.  

### Testing
- All unit tests pass.  
- End-to-end tests pass.  (with Java11, 17, and 21 runtime on Lambda)

### Backward Compatibility
- No risk of breaking existing functionality.  
- The change only adds instrumentation for `RequestStreamHandler` without modifying existing behavior for `RequestHandler`.  
- Existing users who do not use `RequestStreamHandler` remain unaffected.  


srprash:
Hi @lukeina2z . Have you tested this on different Java lambda runtimes (11, 17 and 21) to ensure this works consistently?

PR #1083, opened 2025-05-22
Update dependencies and OTel to v2.11.0-adot1

yiyuan-he:
### What does this pull request do?
Updating patch for `v2.10.0` to `v2.11.0` bump.

Reference for how this patch was created: https://github.com/yiyuan-he/opentelemetry-java-instrumentation/pull/1


instrumentation/logback-1.0/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/logback_1_0/AwsXrayLoggingEventInstrumentation.java
```diff
@@ -26,7 +26,6 @@
 import io.opentelemetry.api.trace.SpanContext;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.instrumentation.api.util.VirtualField;
-import io.opentelemetry.instrumentation.logback.mdc.v1_0.internal.UnionMap;
```
yiyuan-he, line 29: This no longer exists in upstream [v2.11.0](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/v2.11.0/instrumentation/logback/logback-mdc-1.0/library/src/main/java/io/opentelemetry/instrumentation/logback/mdc/v1_0).

[v2.10.0](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/v2.10.0/instrumentation/logback/logback-mdc-1.0/library/src/main/java/io/opentelemetry/instrumentation/logback/mdc/v1_0) for reference.

mxiamxia (dismissed):
All the changes in the patch file are related to updating test cases from Groovy to Java. I reviewed the assertions for the expected AWS attributes, and they remain unchanged. Approving this PR based on this. Thanks

.github/patches/opentelemetry-java-instrumentation.patch
```diff
@@ -1,39 +1,39 @@
...
-index 5c96ee4e2d..9eaed9b327 100644
+index 93437ef1e0..3f564d25bc 100644
 --- a/docs/apidiffs/current_vs_latest/opentelemetry-instrumentation-annotations.txt
 +++ b/docs/apidiffs/current_vs_latest/opentelemetry-instrumentation-annotations.txt
 @@ -1,2 +1,2 @@
--Comparing source compatibility of opentelemetry-instrumentation-annotations-2.10.0.jar against opentelemetry-instrumentation-annotations-2.9.0.jar
-+Comparing source compatibility of opentelemetry-instrumentation-annotations-2.10.0-adot1.jar against opentelemetry-instrumentation-annotations-2.10.0.jar
- No changes.
+ Comparing source compatibility of opentelemetry-instrumentation-annotations-2.11.0.jar against opentelemetry-instrumentation-annotations-2.10.0.jar
+-No changes.
```
thpierce, line 7: Why are we not including the `adot1`

instrumentation/logback-1.0/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/logback_1_0/AwsXrayLoggingEventInstrumentation.java
```diff
@@ -93,7 +92,9 @@ public static void onExit(
       if (contextData == null) {
         contextData = spanContextData;
       } else {
-        contextData = new UnionMap<>(contextData, spanContextData);
+        Map<String, String> mergedData = new HashMap<>(contextData);
+        mergedData.putAll(spanContextData);
+        contextData = mergedData;
```
thpierce, line 97: Seems to be different functionality than Union map. Why do you think this is acceptable?

thpierce: Specifically, the new behaviour is "If the map previously contained a mapping for the key, the old value is replaced by the specified value" so the values of spanContextData overwrite the values of contextData, but in `UnionMap`, I believe the values of spanContextData (first) take priority.

.github/patches/opentelemetry-java-instrumentation.patch
yiyuan-he: We should - updated to include `adot1`.

instrumentation/logback-1.0/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/logback_1_0/AwsXrayLoggingEventInstrumentation.java
yiyuan-he: Thanks for the catch Thomas - updated with changes that should be 1:1 with `UnionMap` functionaltiy.

thpierce (approved):

licenses/annotations-2.29.23.jar/META-INF/LICENSE.txt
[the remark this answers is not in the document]
thpierce: Ack, I skimmed and nothing looked obviously super weird.

mxiamxia (approved):

PR #1085, opened 2025-05-28
Fix: Lambda Topology Issue (#1016)

yiyuan-he:
## What does this pull request do?
Cherry picks the Lambda Topology fix from our Lambda Layer v1 branch.

Original PR: https://github.com/aws-observability/aws-otel-java-instrumentation/pull/1016

## Test strategy
Unit tests and followed e2e test strategy used in the [original PR](https://github.com/aws-observability/aws-otel-java-instrumentation/pull/1016) to sanity check correct behavior in service topology.


awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -518,6 +521,37 @@ private static void setRemoteResourceTypeAndIdentifier(SpanData span, Attributes
...
+          Optional<String> remoteService =
+              getLambdaFunctionNameFromArn(
+                  Optional.ofNullable(escapeDelimiters(span.getAttributes().get(AWS_LAMBDA_NAME))));
+          builder.put(AWS_REMOTE_SERVICE, remoteService.get());
+
+          String remoteEnvironment =
+              Optional.ofNullable(System.getenv("LAMBDA_APPLICATION_SIGNALS_REMOTE_ENVIRONMENT"))
+                  .filter(s -> !s.isEmpty())
+                  .orElse("default");
+          builder.put(AWS_REMOTE_ENVIRONMENT, "lambda:" + remoteEnvironment);
```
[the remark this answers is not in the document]
yiyuan-he: Ack. Will add a test case for this.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -518,6 +521,37 @@ private static void setRemoteResourceTypeAndIdentifier(SpanData span, Attributes
...
+        // https://github.com/aws-observability/aws-otel-python-instrumentation/pull/319
+        //
+        // NOTE: The environment variables LAMBDA_APPLICATION_SIGNALS_REMOTE_ENVIRONMENT was
+        // introduced as part of this fix.
+        // It is optional and allows users to override the default value if needed.
+        if ("Invoke".equals(getRemoteOperation(span, RPC_METHOD))) {
+          Optional<String> remoteService =
+              getLambdaFunctionNameFromArn(
+                  Optional.ofNullable(escapeDelimiters(span.getAttributes().get(AWS_LAMBDA_NAME))));
+          builder.put(AWS_REMOTE_SERVICE, remoteService.get());
```
thpierce, line 540: It is not appropriate to set `AWS_REMOTE_SERVICE` in this method, move to `setRemoteServiceAndOperation` (really `normalizeRemoteServiceName`).

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
thpierce, line 546: Also this should be moved to a new method `setRemoteEnvironment` or MAYBE be tied to `setRemoteServiceAndOperation`. You will need to justify this.

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
```diff
@@ -875,6 +876,11 @@ public void testSdkClientSpanWithRemoteResourceAttributes() {
     validateRemoteResourceAttributes("AWS::SecretsManager::Secret", "secretName");
     mockAttribute(AWS_SECRET_ARN, null);
 
+    // Validate behaviour of AWS_LAMBDA_NAME, then remove it.
+    mockAttribute(AWS_LAMBDA_NAME, "testLambdaName");
+    validateRemoteResourceAttributes("AWS::Lambda::Function", "testLambdaName");
```
thpierce, line 881: Also need to test the invoke case does NOT set resource attributes.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -518,6 +521,37 @@ private static void setRemoteResourceTypeAndIdentifier(SpanData span, Attributes
...
+      } else if (isKeyPresent(span, AWS_LAMBDA_NAME)) {
+        // Handling downstream Lambda as a service vs. an AWS resource:
+        // - If the method call is "Invoke", we treat downstream Lambda as a service.
+        // - Otherwise, we treat it as an AWS resource.
+        //
+        // This addresses a Lambda topology issue in Application Signals.
+        // More context in PR:
+        // https://github.com/aws-observability/aws-otel-python-instrumentation/pull/319
+        //
+        // NOTE: The environment variables LAMBDA_APPLICATION_SIGNALS_REMOTE_ENVIRONMENT was
```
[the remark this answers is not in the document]
thpierce: I also sort of feel this should have been passed through from `AwsApplicationSignalsCustomizerProvider`, as we should have used `ConfigProperties` to extract this as we do for all other env vars.

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
yiyuan-he: Done. The test `testLambdaInvokeSpanWithRemoteServiceAttributes` explicitly verifies that resource attributes are NOT set when Lambda is invoked.

Also added a few more tests for additional coverage:
- `testLambdaInvokeSpanWithLambdaArnInsteadOfName` - tests ARN parsing
- `testLambdaNonInvokeSpanWithRemoteResourceAttributes` - tests non-Invoke as resource
- `testLambdaNonInvokeSpanWithLambdaArnInsteadOfName` - tests non-Invoke with ARN

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
yiyuan-he: Done. Created a dedicated `setRemoteEnvironment` method to keep logic separated and maintain the existing patterns.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -256,6 +260,29 @@ private static void setRemoteServiceAndOperation(SpanData span, AttributesBuilde
...
+          && isKeyPresent(span, AWS_LAMBDA_NAME)
+          && "Invoke".equals(remoteOperation)) {
+        // When invoking Lambda, treat it as a service rather than a resource
+        Optional<String> lambdaFunctionName =
+            getLambdaFunctionNameFromArn(
+                Optional.ofNullable(escapeDelimiters(span.getAttributes().get(AWS_LAMBDA_NAME))));
+        if (lambdaFunctionName.isPresent()) {
+          remoteService = lambdaFunctionName.get();
+        }
+      }
```
thpierce, line 284: IMO, we should move this below to `normalizeRemoteServiceName`. it's confusing to set something then overwrite it when both pierce of logic are hyper-focused on AWS spans:

```
        case "AWSLambda": // AWS SDK v1
        case "Lambda": // AWS SDK v2
            Optional<String> lambdaFunctionName = getLambdaFunctionNameFromArn(....); 
            if (isLambdaInvokeOperation(span) && lambdaFunctionName.isPresent()) {
                return lambdaFunctionName.get();
             } else {
                return NORMALIZED_LAMBDA_SERVICE_NAME;
            }
```

Is there really a scenario where isLambdaInvokeOperation and !lambdaFunctionName.isPresent? It's ok if not, defensive programming is a good idea, but if there is, are we sure we are doing the right thing by setting remote  service to NORMALIZED_LAMBDA_SERVICE_NAME?

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -291,6 +318,30 @@ private static void setRemoteServiceAndOperation(SpanData span, AttributesBuilde
...
+   * Sets the remote environment attribute for special service cases. Currently handles Lambda
+   * invoke operations where the downstream Lambda is treated as a service.
+   *
+   * @param span Span data
+   * @param builder Span attributes builder
+   */
+  private static void setRemoteEnvironment(SpanData span, AttributesBuilder builder) {
+    // Check if this is a Lambda Invoke operation
+    if (isAwsSDKSpan(span)
+        && isKeyPresent(span, AWS_LAMBDA_NAME)
```
thpierce, line 331: Actually I think we should check `getRemoteService(span, RPC_SERVICE)` is "AWSLambda" (AWS SDK v1) or "Lambda" (AWS SDK v2), that's a more reliable check IMO. Also I would re-write this to be more aligned with my comment on `setRemoteServiceAndOperation`, above.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -291,6 +318,30 @@ private static void setRemoteServiceAndOperation(SpanData span, AttributesBuilde
     builder.put(AWS_REMOTE_OPERATION, remoteOperation);
   }
 
+  /**
+   * Sets the remote environment attribute for special service cases. Currently handles Lambda
+   * invoke operations where the downstream Lambda is treated as a service.
+   *
+   * @param span Span data
+   * @param builder Span attributes builder
```
thpierce, line 326: You don't normally add javadoc like this for private methods. Advise ditching it.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -291,6 +318,30 @@ private static void setRemoteServiceAndOperation(SpanData span, AttributesBuilde
...
+   */
+  private static void setRemoteEnvironment(SpanData span, AttributesBuilder builder) {
+    // Check if this is a Lambda Invoke operation
+    if (isAwsSDKSpan(span)
+        && isKeyPresent(span, AWS_LAMBDA_NAME)
+        && "Invoke".equals(span.getAttributes().get(RPC_METHOD))) {
```
thpierce, line 332: Make "Invoke" a constant.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -256,6 +260,29 @@ private static void setRemoteServiceAndOperation(SpanData span, AttributesBuilde
...
+      // - If the method call is "Invoke", we treat downstream Lambda as a service.
+      // - Otherwise, we treat it as an AWS resource.
+      //
+      // This addresses a Lambda topology issue in Application Signals.
+      // More context in PR:
+      // https://github.com/aws-observability/aws-otel-python-instrumentation/pull/319
+      //
+      // NOTE: The environment variable LAMBDA_APPLICATION_SIGNALS_REMOTE_ENVIRONMENT was
+      // introduced as part of this fix.
+      // It is optional and allows users to override the default value if needed.
```
thpierce, line 273: This is overly verbose but also doesn't explain things any better. Amend to:

```
// We want to treat downstream Lambdas as a service rather than a resource because <brief explanation of problem - Application Signals topology map disconnected because ...>
// Additional context can be found in https://github.com/aws-observability/aws-otel-python-instrumentation/pull/319
```

Make `LAMBDA_APPLICATION_SIGNALS_REMOTE_ENVIRONMENT` a constant and add a comment above THAT explaining what it is.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -518,6 +569,17 @@ private static void setRemoteResourceTypeAndIdentifier(SpanData span, Attributes
                 Optional.ofNullable(escapeDelimiters(span.getAttributes().get(AWS_SECRET_ARN))));
         cloudformationPrimaryIdentifier =
             Optional.ofNullable(escapeDelimiters(span.getAttributes().get(AWS_SECRET_ARN)));
+      } else if (isKeyPresent(span, AWS_LAMBDA_NAME)) {
+        // For non-Invoke Lambda operations, treat as a resource
+        // Invoke operations are handled in setRemoteServiceAndOperation
```
thpierce, line 574:
```suggestion
        // For non-Invoke Lambda operations, treat Lambda as a resource, 
        // see setRemoteServiceAndOperation for more information.
```

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -518,6 +569,17 @@ private static void setRemoteResourceTypeAndIdentifier(SpanData span, Attributes
...
+        if (!"Invoke".equals(span.getAttributes().get(RPC_METHOD))) {
+          remoteResourceType = Optional.of(NORMALIZED_LAMBDA_SERVICE_NAME + "::Function");
+          remoteResourceIdentifier =
+              getLambdaFunctionNameFromArn(
```
thpierce, line 578: Why do we call `getLambdaFunctionNameFromArn` on `AWS_LAMBDA_NAME`?  I see we also have `AWS_LAMBDA_ARN`, wouldn't that be an ARN and `AWS_LAMBDA_NAME` be a name?

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
```diff
@@ -750,6 +754,86 @@ public void testPeerServiceDoesNotOverrideAwsRemoteService() {
...
+    // When RPC_METHOD is "Invoke", Lambda should be treated as a service
+    assertThat(actualAttributes.get(AWS_REMOTE_SERVICE)).isEqualTo("myLambdaFunction");
+    assertThat(actualAttributes.get(AWS_REMOTE_ENVIRONMENT)).isEqualTo("lambda:default");
+    // Should not set resource attributes when treated as a service
+    assertThat(actualAttributes.get(AWS_REMOTE_RESOURCE_TYPE)).isNull();
+    assertThat(actualAttributes.get(AWS_REMOTE_RESOURCE_IDENTIFIER)).isNull();
+
+    mockAttribute(RPC_METHOD, null);
+    mockAttribute(AWS_LAMBDA_NAME, null);
+  }
```
thpierce, line 776: Lets try to use existing methods (which are more robust and are used by every other case), like `testSdkClientSpanWithRemoteResourceAttributes`:

```
    // Validate behaviour of AWS_LAMBDA_NAME
    mockAttribute(AWS_LAMBDA_NAME, "lambdaName");
    validateRemoteResourceAttributes("AWS::Lambda::Function", "lambdaName");
    mockAttribute(AWS_LAMBDA_NAME, null);

    // Validate behaviour of AWS_LAMBDA_NAME for Invoke
    mockAttribute(AWS_LAMBDA_NAME, "lambdaName");
    mockAttribute(RPC_METHOD, "Invoke");
    validateRemoteResourceAttributes(null, null); // May need to modify this method.
    mockAttribute(AWS_LAMBDA_NAME, null);
    mockAttribute(RPC_METHOD, null);

    // Validate behaviour of AWS_LAMBDA_RESOURCE_ID
    ...
```

Similar for `testRemoteAttributesCombinations`, you may have to slightly modify the implementation to cover AWS_REMOTE_ENVIRONMENT, but that will just make the tests more robust if/when we add more AWS_REMOTE_ENVIRONMENT behaviour in future.

This goes for all below tests.

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
```diff
@@ -750,6 +754,86 @@ public void testPeerServiceDoesNotOverrideAwsRemoteService() {
...
+
+  @Test
+  public void testLambdaInvokeSpanWithLambdaArnInsteadOfName() {
+    mockAttribute(RPC_SYSTEM, "aws-api");
+    mockAttribute(RPC_METHOD, "Invoke");
+    mockAttribute(
+        AWS_LAMBDA_NAME, "arn:aws:lambda:us-east-1:123456789012:function:myLambdaFunction");
```
thpierce, line 783: Wait so sometimes AWS_LAMBDA_NAME is a name and sometimes its an ARN? even though we have AWS_LAMBDA_ARN?

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
```diff
@@ -750,6 +754,86 @@ public void testPeerServiceDoesNotOverrideAwsRemoteService() {
...
+
+    // When RPC_METHOD is not "Invoke", Lambda should be treated as a resource
+    validateRemoteResourceAttributes("AWS::Lambda::Function", "myLambdaFunction");
+
+    // Also verify cloudformation primary identifier is set to ARN
+    when(spanDataMock.getKind()).thenReturn(SpanKind.CLIENT);
+    Attributes actualAttributes =
+        GENERATOR.generateMetricAttributeMapFromSpan(spanDataMock, resource).get(DEPENDENCY_METRIC);
+    assertThat(actualAttributes.get(AWS_CLOUDFORMATION_PRIMARY_IDENTIFIER))
+        .isEqualTo("arn:aws:lambda:us-east-1:123456789012:function:myLambdaFunction");
```
thpierce, line 813: I'm just seeing that this is the ONLY test testing AWS_CLOUDFORMATION_PRIMARY_IDENTIFIER - that is disappointing, seeing as we have many usages of AWS_CLOUDFORMATION_PRIMARY_IDENTIFIER in the code base. Lets try to meet the bar here and add AWS_CLOUDFORMATION_PRIMARY_IDENTIFIER assertion logic to `testSdkClientSpanWithRemoteResourceAttributes` and not just add it for one single test.

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
```diff
@@ -875,6 +959,25 @@ public void testSdkClientSpanWithRemoteResourceAttributes() {
...
+    mockAttribute(RPC_METHOD, "Invoke");
+    mockAttribute(AWS_LAMBDA_NAME, "testLambdaName");
+    when(spanDataMock.getKind()).thenReturn(SpanKind.CLIENT);
+    Attributes actualAttributes =
+        GENERATOR.generateMetricAttributeMapFromSpan(spanDataMock, resource).get(DEPENDENCY_METRIC);
+    // Lambda Invoke should be treated as a service, not a resource
+    assertThat(actualAttributes.get(AWS_REMOTE_SERVICE)).isEqualTo("testLambdaName");
+    assertThat(actualAttributes.get(AWS_REMOTE_RESOURCE_TYPE)).isNull();
+    assertThat(actualAttributes.get(AWS_REMOTE_RESOURCE_IDENTIFIER)).isNull();
+    assertThat(actualAttributes.get(AWS_CLOUDFORMATION_PRIMARY_IDENTIFIER)).isNull();
```
thpierce, line 977: Absolutely not. Look at the surrounding code and align yours with it.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
yiyuan-he: Removed

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
yiyuan-he: Done:
```
private static final String LAMBDA_INVOKE_OPERATION = "Invoke";
```

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
yiyuan-he: Done
```
// Environment variable to specify remote environment for Lambda services in Application Signals
  // topology
  private static final String LAMBDA_APPLICATION_SIGNALS_REMOTE_ENVIRONMENT =
      "LAMBDA_APPLICATION_SIGNALS_REMOTE_ENVIRONMENT";
```

New amended comment:
```
// We want to treat downstream Lambdas as a service rather than a resource because
    // Application Signals topology map gets disconnected due to conflicting Lambda Entity
    // definitions
    // Additional context can be found in
    // https://github.com/aws-observability/aws-otel-python-instrumentation/pull/319
    if (isLambdaInvokeOperation(span)) {
```

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
yiyuan-he: Discussed offline - added this comment to clarify this to future developers:
```
// AWS_LAMBDA_NAME can contain either a function name or function ARN since Lambda AWS
            // SDK calls accept both formats
            Optional<String> lambdaFunctionName =
                getLambdaFunctionNameFromArn(
                    Optional.ofNullable(
                        escapeDelimiters(span.getAttributes().get(AWS_LAMBDA_NAME))));
```

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
yiyuan-he: Addressed in [thread](https://github.com/aws-observability/aws-otel-java-instrumentation/pull/1085#discussion_r2116817678)

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -373,6 +402,12 @@ private static String generateRemoteService(SpanData span) {
     return remoteService;
   }
 
+  private static boolean isLambdaInvokeOperation(SpanData span) {
```
thpierce, line 405: Technically you should also call and assert `isAwsSDKSpan(span)`

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -420,7 +455,17 @@ private static String normalizeRemoteServiceName(SpanData span, String serviceNa
...
         case "AWSLambda": // AWS SDK v1
         case "Lambda": // AWS SDK v2
-          return NORMALIZED_LAMBDA_SERVICE_NAME;
+          if (isLambdaInvokeOperation(span)) {
+            // AWS_LAMBDA_NAME can contain either a function name or function ARN since Lambda AWS
+            // SDK calls accept both formats
+            Optional<String> lambdaFunctionName =
+                getLambdaFunctionNameFromArn(
+                    Optional.ofNullable(
+                        escapeDelimiters(span.getAttributes().get(AWS_LAMBDA_NAME))));
```
thpierce: Actually I don't know that we do need to escape delimiters, AFAIK only resource identifiers need this. Check if we delimit remote/local environment in CWAgent.

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
```diff
@@ -483,6 +488,31 @@ public void testRemoteAttributesCombinations() {
...
+    validateExpectedRemoteAttributes("testFunction", "Invoke");
+    // Also validate AWS_REMOTE_ENVIRONMENT is set for Lambda Invoke
+    when(spanDataMock.getKind()).thenReturn(SpanKind.CLIENT);
+    Attributes actualAttributes =
+        GENERATOR.generateMetricAttributeMapFromSpan(spanDataMock, resource).get(DEPENDENCY_METRIC);
+    assertThat(actualAttributes.get(AWS_REMOTE_ENVIRONMENT)).isEqualTo("lambda:default");
+    mockAttribute(RPC_SYSTEM, null);
+    mockAttribute(RPC_SERVICE, null);
+    mockAttribute(RPC_METHOD, null);
+    mockAttribute(AWS_LAMBDA_NAME, null);
```
thpierce: This belongs in `testNormalizeRemoteServiceName_AwsSdk`:
```
    testAwsSdkServiceNormalization("Lambda", "AWS::AWS::Lambda");
    mockAttribute(RPC_METHOD, "Invoke");
    mockAttribute(AWS_LAMBDA_NAME, "testFunction");
    testAwsSdkServiceNormalization("Lambda", "testFunction");
    mockAttribute(RPC_METHOD, null);
    mockAttribute(AWS_LAMBDA_NAME, null);

...

    testAwsSdkServiceNormalization("AWSLambda", "AWS::Lambda");
    mockAttribute(RPC_METHOD, "Invoke");
    mockAttribute(AWS_LAMBDA_NAME, "testFunction");
    testAwsSdkServiceNormalization("AWSLambda", "testFunction");
    mockAttribute(RPC_METHOD, null);
    mockAttribute(AWS_LAMBDA_NAME, null);
```

Move remote enviroment logic to it's own test `testSetRemoteEnvironment` where you show that only by setting all relevant attributes can you get remote enviroment to be set, and missing any causes remote enviroment to be unset.

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
```diff
@@ -839,6 +869,17 @@ public void testSdkClientSpanWithRemoteResourceAttributes() {
...
+    mockAttribute(AWS_GUARDRAIL_ID, "test_guardrail_id");
+    mockAttribute(
+        AWS_GUARDRAIL_ARN, "arn:aws:bedrock:us-east-1:123456789012:guardrail/test_guardrail_id");
+    validateRemoteResourceAttributes(
+        "AWS::Bedrock::Guardrail",
+        "test_guardrail_id",
+        "arn:aws:bedrock:us-east-1:123456789012:guardrail/test_guardrail_id");
+    mockAttribute(AWS_GUARDRAIL_ID, null);
+    mockAttribute(AWS_GUARDRAIL_ARN, null);
+
```
thpierce: I would just merge this into above two tests (so we also test delimiters in cloudformationPrimaryIdentifier, not sure if we need a separate test.

Though what we do need is a test where `cloudformationPrimaryIdentifier` is not present, and is replaced with `remoteResourceIdentifier` in the final part of `setRemoteResourceTypeAndIdentifier`

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
```diff
@@ -854,27 +895,83 @@ public void testSdkClientSpanWithRemoteResourceAttributes() {
     mockAttribute(
```
thpierce, line 876: Starting with this part of the tests, we stopped testing attribute with special chars(^). You don't need to fix this in this PR, but please cut a backlog item to add these tests.

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
```diff
@@ -854,27 +895,83 @@ public void testSdkClientSpanWithRemoteResourceAttributes() {
...
+    // Validate that Lambda Invoke with function name treats Lambda as a service, not a resource
+    mockAttribute(RPC_SERVICE, "Lambda");
+    mockAttribute(RPC_METHOD, "Invoke");
+    mockAttribute(AWS_LAMBDA_NAME, "testLambdaName");
+    validateRemoteResourceAttributes(null, null);
+    mockAttribute(RPC_SERVICE, null);
+    mockAttribute(RPC_METHOD, null);
+    mockAttribute(AWS_LAMBDA_NAME, null);
+
+    // Validate that Lambda Invoke with ARN in AWS_LAMBDA_NAME extracts function name correctly
```
thpierce: Inaccurate comment. Consider merging into above comment `// Validate behaviour of AWS_LAMBDA_NAME containing ARN for non-Invoke operations`

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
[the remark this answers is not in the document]
thpierce: Thread: https://github.com/aws-observability/aws-otel-java-instrumentation/pull/1085#discussion_r2114553216

Lets not let this block this PR, can discuss offline. 

However, let's move the `LAMBDA_APPLICATION_SIGNALS_REMOTE_ENVIRONMENT` constant to `AwsApplicationSignalsCustomizerProvider`, similar to `AWS_LAMBDA_FUNCTION_NAME`

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -420,7 +455,17 @@ private static String normalizeRemoteServiceName(SpanData span, String serviceNa
...
+            return lambdaFunctionName.get();
```
thpierce: Hm - what if AWS_LAMBDA_NAME was not present? We would just use `UnknownRemoteService`? Just want to make sure that's an intentional decision. Lets also add a test for this case.

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
yiyuan-he: Created a backlog item for this

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
yiyuan-he: Followed-up offline. After reviewing the O2F spec sheet it seems like we don't do any special processing for remote/local environment.

Made this change to remove the `escapeDelimiters` call:
```
            Optional<String> lambdaFunctionName =
                getLambdaFunctionNameFromArn(
                    Optional.ofNullable(span.getAttributes().get(AWS_LAMBDA_NAME)));
```

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
yiyuan-he: Done. Moved the the above logic to `testNormalizeRemoteServiceName_AwsSdk`. Also added a new `testSetRemoteEnvironment` test case that covers the following scenarios:
- Setting remote environment when all attributes are present
- NOT setting it when `RPC_SYSTEM` is missing
- NOT setting it when `RPC_METHOD` is missing
- Still setting it to `lambda:default` when `AWS_LAMBDA_NAME` is missing (discussed offline - this is an intentional fallback behavior)
- NOT setting it for non-Lambda services

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
yiyuan-he: Done. Added this assertion at the beginning of `isLambdaInvokeOperation`:
```
  private static boolean isLambdaInvokeOperation(SpanData span) {
    if (!isAwsSDKSpan(span)) {
      return false;
    }
    String rpcService = getRemoteService(span, RPC_SERVICE);
    return ("AWSLambda".equals(rpcService) || "Lambda".equals(rpcService))
        && LAMBDA_INVOKE_OPERATION.equals(span.getAttributes().get(RPC_METHOD));
  }
```

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
yiyuan-he: Done. Merged this into the existing two tests above. Also created a new `testCloudFormationPrimaryIdentifierFallbackToRemoteResourceIdentifier` method to test fallback cases. I've tried to cover most of the fallback scenarios, will defer any additional cases to the backlog task.

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
yiyuan-he: Done. Updated comment to:
```
    mockAttribute(RPC_METHOD, null);
    mockAttribute(AWS_LAMBDA_NAME, null);

    // Validate behaviour of AWS_LAMBDA_NAME containing ARN for Invoke operations
```

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
yiyuan-he: Discussed offline - we make an intentional decision to fallback to `UnknownRemoteService` here.

Updated implementation to reflect this:
```
            return lambdaFunctionName.orElse(UNKNOWN_REMOTE_SERVICE);
```

Added the follow to our tests to cover this fallback behavior:
```
    // Test Lambda Invoke without AWS_LAMBDA_NAME - should fall back to UnknownRemoteService
    mockAttribute(AWS_LAMBDA_NAME, null);
    testAwsSdkServiceNormalization("Lambda", "UnknownRemoteService");
    mockAttribute(RPC_METHOD, null);

    testAwsSdkServiceNormalization("AWSLambda", "AWS::Lambda");
    mockAttribute(RPC_METHOD, "Invoke");
    mockAttribute(AWS_LAMBDA_NAME, "testFunction");
    testAwsSdkServiceNormalization("AWSLambda", "testFunction");
    // Test Lambda Invoke without AWS_LAMBDA_NAME - should fall back to UnknownRemoteService
    mockAttribute(AWS_LAMBDA_NAME, null);
    testAwsSdkServiceNormalization("AWSLambda", "UnknownRemoteService");
    mockAttribute(RPC_METHOD, null);

```

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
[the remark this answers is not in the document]
yiyuan-he: Sounds good - discussed offline.

Moved this env var to `AwsApplicationSignalsCustomizerProvider` for better organization. 

It is now consumed in `AwsMetricAttributeGenerator.java` via import:
```
import static software.amazon.opentelemetry.javaagent.providers.AwsApplicationSignalsCustomizerProvider.LAMBDA_APPLICATION_SIGNALS_REMOTE_ENVIRONMENT;
```

Will treat passing in as `ConfigProperties` as lower priority follow-up since there exist some counter-examples in the codebase.

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
```diff
@@ -1417,6 +1506,61 @@ private void testAwsSdkServiceNormalization(String serviceName, String expectedR
...
+    mockAttribute(RPC_METHOD, "GetFunction");
+    actualAttributes =
+        GENERATOR.generateMetricAttributeMapFromSpan(spanDataMock, resource).get(DEPENDENCY_METRIC);
+    assertThat(actualAttributes.get(AWS_REMOTE_ENVIRONMENT)).isNull();
+
+    // Clean up
+    mockAttribute(RPC_SYSTEM, null);
+    mockAttribute(RPC_SERVICE, null);
+    mockAttribute(RPC_METHOD, null);
+    mockAttribute(AWS_LAMBDA_NAME, null);
```
thpierce: Not blocking: I don't think you need these, since attributesMock is recreated `BeforeEach` test.

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
```diff
@@ -1417,6 +1506,61 @@ private void testAwsSdkServiceNormalization(String serviceName, String expectedR
     assertThat(actualAttributes.get(AWS_REMOTE_SERVICE)).isEqualTo(expectedRemoteService);
   }
 
+  @Test
+  public void testSetRemoteEnvironment() {
```
thpierce, line 1479: Not blocking: I may have refactored this slightly:

```
public void testSetRemoteEnvironment() {
   validateLambdaRemoteEnvironment("aws-api", "Lambda", "Invoke", "testFunction", "lambda:default");
   validateLambdaRemoteEnvironment(null, "Lambda", "Invoke", "testFunction", null);
   validateLambdaRemoteEnvironment("aws-api",null, "Invoke", "testFunction", null);
   validateLambdaRemoteEnvironment("aws-api", "Lambda", null, "testFunction", null);
   validateLambdaRemoteEnvironment("aws-api", "Lambda", "Invoke", null, null);
   ...
}
```

And then within validateLambdaRemoteEnvironment you could have done the same SpanKind testing logic as other tests (if SpanKind==CLIENT, metric produced, if ==SERVER, no metric produced), etc.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -291,6 +299,23 @@ private static void setRemoteServiceAndOperation(SpanData span, AttributesBuilde
     builder.put(AWS_REMOTE_OPERATION, remoteOperation);
   }
 
+  private static void setRemoteEnvironment(SpanData span, AttributesBuilder builder) {
```
thpierce, line 306: Not blocking: All other `set*` methods have a comment summarizing the behaviour, consider aligning.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -518,6 +564,19 @@ private static void setRemoteResourceTypeAndIdentifier(SpanData span, Attributes
                 Optional.ofNullable(escapeDelimiters(span.getAttributes().get(AWS_SECRET_ARN))));
         cloudformationPrimaryIdentifier =
             Optional.ofNullable(escapeDelimiters(span.getAttributes().get(AWS_SECRET_ARN)));
+      } else if (isKeyPresent(span, AWS_LAMBDA_NAME)) {
+        // For non-Invoke Lambda operations, treat Lambda as a resource,
+        // see normalizeRemoteServiceName for more information.
+        if (!LAMBDA_INVOKE_OPERATION.equals(span.getAttributes().get(RPC_METHOD))) {
```
thpierce: Not blocking: Would it be better to call `isLambdaInvokeOperation` here?

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
```diff
@@ -883,6 +961,95 @@ public void testSdkClientSpanWithRemoteResourceAttributes() {
     mockAttribute(RPC_SYSTEM, "null");
   }
 
+  @Test
+  public void testCloudFormationPrimaryIdentifierFallbackToRemoteResourceIdentifier() {
```
thpierce, line 965: Not blocking: Could probably refactor this to be a bit easier to parse.

```
public void testCloudFormationPrimaryIdentifierFallbackToRemoteResourceIdentifier() {
    mockAttribute(RPC_SYSTEM, "aws-api");
    when(spanDataMock.getKind()).thenReturn(SpanKind.CLIENT);

    // Test case 1: S3 Bucket (no ARN available, should use bucket name for both)
    mockAttribute(AWS_BUCKET_NAME, "my-test-bucket");
    validateRemoteResourceAttributes("AWS::S3::Bucket", "my-test-bucket")

    ...
}
```

thpierce: Not blocking: You did miss the Lambda case here. Ok if we want to take as backlog item.

thpierce (approved):

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
yiyuan-he: Added this docstring:
```
/**
   * Remote environment is used to identify the environment of downstream services.
   * Currently only set to "lambda:default" for Lambda Invoke operations when aws-api system is
   * detected.
   */
```
Tried to leave it more general which leaves room for future expansion but also callout that it's just being used for Lambda currently.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
yiyuan-he: Done. Changed to `isLambdaInvokeOperation`.

```
      } else if (isKeyPresent(span, AWS_LAMBDA_NAME)) {
        // For non-Invoke Lambda operations, treat Lambda as a resource,
        // see normalizeRemoteServiceName for more information.
        if (!isLambdaInvokeOperation(span)) {
```

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
yiyuan-he: Good point - I should have just reused `validateRemoteResurceAttributes`.

Updated with this refactor and added the Lambda case as well.

```
  @Test
  public void testCloudFormationPrimaryIdentifierFallbackToRemoteResourceIdentifier() {
    // Test that when cloudformationPrimaryIdentifier is not explicitly set,
    // it falls back to use the same value as remoteResourceIdentifier

    mockAttribute(RPC_SYSTEM, "aws-api");
    when(spanDataMock.getKind()).thenReturn(SpanKind.CLIENT);

    // Test case 1: S3 Bucket (no ARN available, should use bucket name for both)
    mockAttribute(AWS_BUCKET_NAME, "my-test-bucket");
    validateRemoteResourceAttributes("AWS::S3::Bucket", "my-test-bucket");

    // Test S3 bucket with special characters
    mockAttribute(AWS_BUCKET_NAME, "my-test|bucket^name");
    validateRemoteResourceAttributes("AWS::S3::Bucket", "my-test^|bucket^^name");
    mockAttribute(AWS_BUCKET_NAME, null);

    ...

    // Test case 5: Lambda Function (non-invoke operation, no ARN)
    mockAttribute(RPC_METHOD, "GetFunction"); // Non-invoke operation
    mockAttribute(AWS_LAMBDA_NAME, "my-test-function");
    validateRemoteResourceAttributes("AWS::Lambda::Function", "my-test-function");

    // Test Lambda function with special characters
    mockAttribute(AWS_LAMBDA_NAME, "my-function|with^chars");
    validateRemoteResourceAttributes("AWS::Lambda::Function", "my-function^|with^^chars");
    mockAttribute(AWS_LAMBDA_NAME, null);
    mockAttribute(RPC_METHOD, null);
```

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
yiyuan-he: Was not able to get to this in timebox. Will defer to the backlog item

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
yiyuan-he: Done. Removed this unecessary attribute clean up.

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
thpierce: Fair enough.

thpierce (approved):

PR #1089, opened 2025-05-29
Added patch for the X-Ray Remote Sampler

AsakerMohd:
### Description of changes:
Since upstream has updated the HTTP semantic conventions, the X-Ray Remote Sampler has been outdated as it's been looking at the older attributes that are no longer emitted. I've created a patch to temporarily fix this until the change is merged upstream. 

To create the patch file:
1. Cloned the [opentelemetry-java-contrib](https://github.com/open-telemetry/opentelemetry-java-contrib/) repo
2. Ran `git checkout v1.39.0` to checkout to the same version used by the latest release (2.11.0)
3. Made the changes in the `SamplingRuleApplier.java`, `SamplingRuleApplierTest.java` and the `version.gradle.kts` files
4. Generated the patch file by running `git diff > opentelemetry-java-contrib.patch`. This should be a temporary patch. I'm creating another CR in upstream to fix this issue. Once it's fixed, we will need to update the [io.opentelemetry.contrib:opentelemetry-aws-xray](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/284eed6a7e50237b2a677258bef58e8bf01c71d9/dependencyManagement/build.gradle.kts#L76) version to the latest one.

### Testing:

**Testing Setup:**
1. Created a local build of the ADOT Distro by running the following:
```
./gradlew clean
./scripts/local_patch.sh //applies the patches including the new contrib patch
./gradlew build //this builds everything under the adot repo including the sample apps. 
```
2. Used the [SpringBoot](https://github.com/aws-observability/aws-otel-java-instrumentation/tree/284eed6a7e50237b2a677258bef58e8bf01c71d9/sample-apps/springboot) sample app which includes both the `outgoing-http-call` and the `/aws-sdk-call` APIs
3. Started the AWS Adot Collector in a docker container to retrieve Sampling Rules and to export traces to X-Ray
4. Started the Sample App using the following:
```
JAVA_TOOL_OPTIONS=" -javaagent:/Volumes/workplace/otel/aws-otel-java-instrumentation/otelagent/build/libs/aws-opentelemetry-agent-2.11.0-SNAPSHOT.jar" \
OTEL_METRICS_EXPORTER=none \
OTEL_LOGS_EXPORTER=none \
OTEL_AWS_APPLICATION_SIGNALS_ENABLED=true \
OTEL_AWS_APPLICATION_SIGNALS_EXPORTER_ENDPOINT=http://localhost:4316/v1/metrics \
OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf \
OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://localhost:4316/v1/traces \
OTEL_RESOURCE_ATTRIBUTES="service.name=$YOUR_SVC_NAME" \
java -jar /Volumes/workplace/otel/aws-otel-java-instrumentation/sample-apps/springboot/build/libs/springboot-2.11.0-SNAPSHOT.jar
```

**Manual Testing:**

For the manual testing. I created three sampling rules
1. Default one with zero sampling
6. Second one with path = `/aws-sdk-call` with 100% sampling
7. Third one with path = `/outgoing-http-call` with 100% sampling

Before the change, calling any of the above APIs wasn't being matched when it should. After the change, the rules were being matched and each rule showed the correct stats. 

**Before the change:**
Screenshot 2025-06-01 at 11 34 56 PM

**After the change:**
Screenshot 2025-06-01 at 11 29 54 PM


thpierce (commented):
We need testing showing that this patch can takes effect and is working as expected.

Also please be more verbose in exactly how you generated this patch in the overview.

.github/patches/opentelemetry-java-contrib.patch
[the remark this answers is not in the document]
AsakerMohd: Done

.github/patches/versions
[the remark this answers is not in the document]
AsakerMohd: Added new logic in the build-layer.sh script to patch contrib as well.

thpierce (commented):
Quick thought: Do we need to do this in other languages?

AsakerMohd:
> Quick thought: Do we need to do this in other languages?

I checked the other languages and we are up to date and we don't have an issue there.

lambda-layer/build-layer.sh
```diff
@@ -34,6 +34,22 @@ patch -p1 < "$SOURCEDIR"/patches/StreamHandlerInstrumentation.patch
 popd
 rm -rf opentelemetry-java-instrumentation
 
+contrib_version=$(awk -F'=v' '/OTEL_JAVA_CONTRIB_VERSION/ {print $2}' "$file")
```
srprash, line 37: Does remote sampling work in Lambda? I don't think it does. 
In that case we don't need to apply this patch for lambda layer. It then saves us from adding and maintaining the `OTEL_JAVA_CONTRIB_VERSION` config which can easily fall out of sync with the actual dependency versions.

.github/patches/opentelemetry-java-contrib.patch
srprash, line 1: Does any patch file under this directory gets automatically applied? Have we verified that our build and release processes apply this patch file?

.github/patches/versions
```diff
@@ -1 +1,2 @@
 OTEL_JAVA_INSTRUMENTATION_VERSION=v2.11.0
+OTEL_JAVA_CONTRIB_VERSION=v1.39.0
```
srprash, line 2: Does this version get used in defining the dependencies from the contrib repo? Can we reuse this so that we don't have to maintain the version number in multiple places?

lambda-layer/build-layer.sh
AsakerMohd: We don't need to maintain `OTEL_JAVA_CONTRIB_VERSION` tho. If it's not there, it won't be applied and it'll pull the original version in gradle. Actually, we have to have this here since we updated build.gradle to look at 1.39.0-adot1 which it won't find in maven unless in maven local. And to get it in maven local, we need the build-layer.sh to publish. It also only affects the aws-xray dep in contrib and not others since it's the only that pulls the adot1 version.

.github/patches/opentelemetry-java-contrib.patch
AsakerMohd: Yes. In [release](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/284eed6a7e50237b2a677258bef58e8bf01c71d9/.github/workflows/release-build.yml#L33) and in [main](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/284eed6a7e50237b2a677258bef58e8bf01c71d9/.github/workflows/main-build.yml#L71)

.github/patches/versions
AsakerMohd: This is used to pull the repo using the release tag. In the build.gradle.kts file, it's hardcoded with the adot1 suffix since we override the version in the patch. We could update build.gradle to read from the version file. Lets do it as a follow up tho.

lambda-layer/build-layer.sh
srprash: On second thought, I think it's okay to apply this patch to lambda layer since we may ave more patches in this file which are not just sampler related. It's good idea to keep the patching consistent.

srprash (approved):

PR #1101, opened 2025-06-11
Sigv4 - Add Missing STS Dependency

liustve:
*Issue #, if available:*
Current Sigv4 Exporter is missing the required dependency from AWS SDK library to allow STS AssumeRole from the DefaultCredentialsProvider.

For example, this is the error that pops up if the exporter is used in an EKS environment:
``` WebIdentityTokenCredentialsProvider(): To use web identity tokens, the 'sts' service module must be on the class path. ```

Leading to the credentials provider to default to the worker node role:

```
 WARN io.opentelemetry.exporter.internal.http.HttpExporter - Failed to export spans. Server responded with HTTP status code 403. Error message: User: arn:aws:sts::571600841604:assumed-role/eksctl-spring-demo-cluster-nodegro-NodeInstanceRole-qe5bSShjEHDp/i-0e398f24217b99394 is not authorized to perform: xray:PutTraceSegments because no identity-based policy allows the xray:PutTraceSegments action
 ```

*Description of changes:*

Add STS as a dependency for the OtlpAws Exporters in order to allow the exporter to AssumeRole.
- Bumping dependency to 2.30.17 as that is the version that a major CVE was patched in AWS SDK v2: https://github.com/aws/aws-sdk-java-v2/issues/6098

*Testing*
A sample EKS application was deployed with a custom ADOT Java Agent artifact that has the STS dependency added. I see spans appearing in the ```aws/spans``` log group:

```
{
    "resource": {
        "attributes": {
            "telemetry.distro.version": "2.11.0-aws-SNAPSHOT",
            "host.image.id": "ami-0335cf904e4303efa",
            "process.command_args": [
                "/usr/lib/jvm/java-17-amazon-corretto/bin/java",
                "-Xms512m",
                "-Xmx1024m",
                "-Dspring.profiles.active=prod",
                "-javaagent:/app/opentelemetry-javaagent.jar",
                "-jar",
                "app.jar"
            ],
            "process.runtime.version": "17.0.15+6-LTS",
            "os.type": "linux",
            "process.pid": 1,
            "host.type": "t3.medium",
            "cloud.availability_zone": "us-east-1a",
            "telemetry.sdk.name": "opentelemetry",
            "telemetry.sdk.language": "java",
            "process.runtime.name": "OpenJDK Runtime Environment",
            "service.instance.id": "afe04022-e9c0-4c77-b0a1-0d790fc0200e",
            "os.description": "Linux 6.1.134-152.225.amzn2023.x86_64",
            "host.arch": "amd64",
            "host.name": "ip-192-168-23-12.ec2.internal",
            "telemetry.sdk.version": "1.45.0",
            "cloud.platform": "aws_ec2",
            "host.id": "i-0c88a3c68d45dfd96",
            "deployment.environment": "prod",
            "cloud.region": "us-east-1",
            "service.name": "demo-app",
            "telemetry.distro.name": "opentelemetry-java-instrumentation",
            "cloud.provider": "aws",
            "service.version": "0.0.1-SNAPSHOT",
            "cloud.account.id": "571600841604",
            "process.executable.path": "/usr/lib/jvm/java-17-amazon-corretto/bin/java",
            "process.runtime.description": "Amazon.com Inc. OpenJDK 64-Bit Server VM 17.0.15+6-LTS"
        }
    },
    "scope": {
        "name": "io.opentelemetry.tomcat-10.0",
        "version": "2.11.0-adot1-alpha"
    },
    "traceId": "6849fcc297c5ca4699ec920918d750ee",
    "spanId": "36e360cfd4490b85",
    "flags": 257,
    "name": "GET /hello",
    "kind": "SERVER",
    "startTimeUnixNano": 1749679298293950299,
    "endTimeUnixNano": 1749679298295162110,
    "durationNano": 1211811,
    "attributes": {
        "user_agent.original": "kube-probe/1.32+",
        "aws.local.service": "demo-app",
        "telemetry.extended": "true",
        "network.protocol.version": "1.1",
        "network.peer.port": 35512,
        "url.scheme": "http",
        "thread.name": "http-nio-8080-exec-3",
        "aws.local.environment": "prod",
        "server.address": "192.168.21.181",
        "client.address": "192.168.23.12",
        "network.peer.address": "192.168.23.12",
        "aws.local.operation": "GET /hello",
        "http.status_code": 200,
        "aws.span.kind": "LOCAL_ROOT",
        "url.path": "/hello",
        "http.request.method": "GET",
        "http.route": "/hello",
        "server.port": 8080,
        "PlatformType": "AWS::EC2",
        "http.response.status_code": 200,
        "thread.id": 26
    },
    "status": {
        "code": "UNSET"
    }
}

```


liustve:
Bumping dependency of AWS SDK due to CVE scan failure:
https://github.com/aws-observability/aws-otel-java-instrumentation/actions/runs/15596667877/job/44536199367

liustve:
It's safe to include these AWS SDK dependencies in the ADOT Java Agent because we employ Java class shading during the build process. This technique relocates these classes to a different package namespace at runtime, creating a separate classpath that won't conflict with any AWS SDK dependencies that might be present in the main application. This prevents version conflicts and classloader issues that would otherwise occur when the agent and application use different versions of the same libraries.

Unlike other language implementations where lazy-loading is used and dependencies are checked at runtime, Java requires special consideration. In other languages, the SDKs typically include all necessary packages for STS and authentication, and the code checks if the user has the correct dependencies before loading them. For Java, we need to explicitly import AWS SDK components and perform extensive testing to ensure compatibility, as these dependencies are bundled directly
with our agent rather than being conditionally loaded from the user's environment.

PR #1106, opened 2025-06-27
update local operation of lambda span based on span attribute

pxaws:
*Issue #, if available:*

*Description of changes:*

Currently the local operation of lambda span is hardcoded to function_name//FunctionHandler. In some use cases, some server is running inside lambda function and we should allow the setting of the local operation dynamically at run time. For example, users can use span api to set the attribute `aws.lambda.local.operation.override` and then this value can be used for the local operation. We have an Amazon-internal framework instrumentation that depends on this.

### DIFF
--- a/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsAttributeKeys.java
+++ b/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsAttributeKeys.java
@@ -40,7 +40,7 @@ final class AwsAttributeKeys {
    *       "MyService/handleRequest");
    */
   static final AttributeKey<String> AWS_LAMBDA_LOCAL_OPERATION_OVERRIDE =
-          AttributeKey.stringKey("aws.lambda.local.operation.override");
+      AttributeKey.stringKey("aws.lambda.local.operation.override");
 
   static final AttributeKey<String> AWS_REMOTE_SERVICE =
       AttributeKey.stringKey("aws.remote.service");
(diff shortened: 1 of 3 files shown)


awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsAttributeKeys.java
```diff
@@ -29,6 +29,9 @@ private AwsAttributeKeys() {}
   static final AttributeKey<String> AWS_LOCAL_OPERATION =
       AttributeKey.stringKey("aws.local.operation");
 
+  static final AttributeKey<String> AWS_LAMBDA_LOCAL_OPERATION_OVERRIDE =
+          AttributeKey.stringKey("aws.lambda.local.operation.override");
```
mxiamxia: Pls add the comment why this attribute is needed in what scenarios

pxaws: Done.

mxiamxia (approved):

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsSpanProcessingUtil.java
```diff
@@ -91,6 +92,22 @@ static List<String> getDialectKeywords() {
...
+       *   Span.current().setAttribute(
+       *       "aws.lambda.local.operation.override",
+       *       "MyServiceOperation");
+       *
+       * The code below will detect that override and use it instead of the default.
+       */
+      String operationOverride = span.getAttributes().get(AWS_LAMBDA_LOCAL_OPERATION_OVERRIDE);
+      if (operationOverride != null) {
+        return operationOverride;
+      }
```
srprash, line 110: Can you a few unit tests to cover this change?

pxaws: Done.

bjrara (approved):

mxiamxia (approved):

PR #1111, opened 2025-07-02
Base of AWS SDK v2.2 SPI Implementation

anahatAWS:
Note: this is not the complete SPI implementation 

### Issue
The current ADOT Java SDK implementation relies on a combination of OpenTelemetry SPI and Git patches to extend the OTel SDK functionality. This approach presents several challenges:

- Reduced modularity and maintainability
- Increased risk of errors during OTel SDK version upgrades
- Manual intervention required for patch management
- Limited ecosystem compatibility with upstream OpenTelemetry
- Difficulty in extending functionality for users

This is the skeleton set up for the SPI, which aims to remove the [patch](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/patches/opentelemetry-java-instrumentation.patch) for aws-sdk v2.2 by using OTel's InstrumentationModule SPI extension. This instrumentation is essentially complementing the upstream java agent. It is completely separate from upstream and instruments after the Otel agent.

### Description of Changes
This PR sets up the foundational structure for AWS SDK v2.2 Instrumentation, moving away from the current patching approach. It doesn't modify current ADOT functionality or the upstream span. It just registers the ADOT SPI  implementation and sets up the interceptor hooks.

#### Core Components

1. **AdotAwsSdkInstrumentationModule**
   - Extends OpenTelemetry's `InstrumentationModule` SPI
   - Registers custom interceptors in specific order:
     1. Upstream AWS SDK execution interceptor
     2. ADOT custom interceptor
   - Ensures proper instrumentation sequencing through careful resource registration

2. **AdotTracingExecutionInterceptor**
   - Extends AWS SDK's `ExecutionInterceptor`
   - Hooks into key SDK lifecycle points:
     - `beforeTransmission`: Captures final SDK request after upstream modifications
     - `modifyResponse`: Processes response before span closure in upstream
   - Will be used to enriches spans
   - Acts as central coordinator for all the awssdk_v2_2 components

3. **Resources Folder**
    -  Registers the AdotAwsSdkInstrumentationModule into OTel's SPI extension classpath in META-INF/services
    - Registers AdotTracingExecutionInterceptor into AWS SDK's interceptor classpath in software.amazon.awssdk.global.handlers

### Key Design Decisions
1. **Instrumentation Ordering**
   - Deliberately structured to run after upstream OTel agent
   - Ensures all upstream modifications are captured
   - Maintains compatibility with existing instrumentation

2. **Lifecycle Hook Points**
[...]

PR #1113, opened 2025-07-08
AWS SDK v2.2 SPI Patch Migration

anahatAWS:
Note: This is a continuation of #1111

#### Description of Changes
This implementation builds on the foundation established in PR #1111, transforming the structural setup into a fully functional SPI-based solution that will replace our current patching approach. This PR does not change the current ADOT functionality because patches have not been removed.

The next/final PR for v2.2 will remove the patches for aws-sdk-2.2 and have unit tests to ensure correct SPI functionality and behaviour. The final PR will also pass all the contract-tests once patches are removed.

#### Changes include:
- Migration of patched files into proper package structure:
NOTE: We are not copying entire files from upstream. Instead, we only migrated the new components that were added by our patches and the methods that use these AWS-specific components. I deliberately removed any code that was untouched by our patches to avoid duplicating upstream instrumentation code. This selective migration ensures we maintain only our AWS-specific additions while letting OTel handle its base functionality.
   - `AwsExperimentalAttributes` - [patch](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/patches/opentelemetry-java-instrumentation.patch#L2167) creates new class
   - `AwsSdkRequest` - [patch](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/patches/opentelemetry-java-instrumentation.patch#L2246) on [this otel file](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.11.x/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/internal/AwsSdkRequest.java)
   - `AwsSdkRequestType` - [patch](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/patches/opentelemetry-java-instrumentation.patch#L2324) on [this otel file](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.11.x/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/internal/AwsSdkRequestType.java)
   - `BedrockJsonParser` - [patch](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/patches/opentelemetry-java-instrumentation.patch#L2427) creates new class
   - `FieldMapper` - [patch](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/patches/opentelemetry-java-instrumentation.patch#L2712) on [this otel file](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.11.x/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/internal/FieldMapper.java)
   - `Serializer` - [patch](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/patches/opentelemetry-java-instrumentation.patch#L2731) on [this otel file](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.11.x/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/internal/Serializer.java)
   - `BedrockJsonParserTest` - [patch](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/patches/opentelemetry-java-instrumentation.patch#L2996) creates new class
   - `AbstractAws2ClientTest` - [patch](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/patches/opentelemetry-java-instrumentation.patch#L3123) on [this otel file](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.11.x/instrumentation/aws-sdk/aws-sdk-2.2/testing/src/main/groovy/io/opentelemetry/instrumentation/awssdk/v2_2/AbstractAws2ClientTest.groovy)
- Setup of dependent files directly copied from upstream aws-sdk:
  - `MethodHandleFactory` - copy-pasted from [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.11.x/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/internal/MethodHandleFactory.java)
  - `FieldMapping` - copy-pasted from [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.11.x/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/internal/FieldMapping.java)
  - `AwsJsonProtocolFactoryAccess` - copy-pasted from [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.11.x/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/internal/AwsJsonProtocolFactoryAccess.java)

The classes we copied from upstream are just helper utilities that make it easier to inject span attributes during instrumentation. They're not core functionality that needs to stay in sync with upstream changes, rather standalone utilities that support our simpler, independent instrumentation without creating version lock-in. They're independent of OTel's core functionality, so we don't face the same version dependency issues we had with patching. We're just following upstream's structure for consistency.

These added files:
- Access and modify span attributes
- Provide consistent formatting tools for span attributes
- Can be updated at our convenience if needed

#### Testing
- Existing functionality verified
- Contract tests passing
- Build successful

#### Related
- Skeleton PR for aws-sdk v2.2: https://github.com/aws-observability/aws-otel-java-instrumentation/pull/1111
- Replaces patch: [current patch](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/patches/opentelemetry-java-instrumentation.patch)

### DIFF
--- a/instrumentation/aws-sdk/build.gradle.kts
+++ b/instrumentation/aws-sdk/build.gradle.kts
@@ -16,6 +16,7 @@
 plugins {
   java
   id("com.gradleup.shadow")
+  id("groovy")
 }
 
 base.archivesBaseName = "aws-instrumentation-aws-sdk"
@@ -23,5 +24,11 @@ base.archivesBaseName = "aws-instrumentation-aws-sdk"
 dependencies {
   compileOnly("io.opentelemetry.javaagent:opentelemetry-javaagent-extension-api")
   compileOnly("software.amazon.awssdk:aws-core:2.2.0")
+  compileOnly("software.amazon.awssdk:secretsmanager:2.2.0")
+
   compileOnly("net.bytebuddy:byte-buddy")
+  compileOnly("com.google.code.findbugs:jsr305:3.0.2")
+
+  testImplementation("com.google.guava:guava")
+  testImplementation("io.opentelemetry.javaagent:opentelemetry-testing-common")
 }
--- a/instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkInstrumentationModule.java
+++ b/instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkInstrumentationModule.java
@@ -43,7 +43,23 @@ public class AdotAwsSdkInstrumentationModule extends InstrumentationModule {
   @Override
   public List<String> getAdditionalHelperClassNames() {
     return Arrays.asList(
-        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AdotAwsSdkTracingExecutionInterceptor");
+        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AdotAwsSdkTracingExecutionInterceptor",
+        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AdotTracingExecutionInterceptor$RequestSpanFinisher",
+        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsSdkRequest",
+        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsSdkRequestType",
+        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.FieldMapper",
+        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.FieldMapping",
+        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.FieldMapping$Type",
+        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsExperimentalAttributes",
+        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.BedrockJsonParser",
+        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.BedrockJsonParser$JsonPathResolver",
+        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.BedrockJsonParser$LlmJson",
+        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.BedrockJsonParser$JsonParser",
+        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.MethodHandleFactory",
+        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.MethodHandleFactory$1",
+        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.Serializer",
+        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsJsonProtocolFactoryAccess",
+        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsSdkRequestType$AttributeKeys");
   }
 
   /**
--- a/instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
+++ b/instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
@@ -15,16 +15,48 @@
 
 package software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2;
 
+import static software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsExperimentalAttributes.GEN_AI_SYSTEM;
+import static software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsSdkRequestType.BEDROCKRUNTIME;
+
+import io.opentelemetry.api.trace.Span;
+import software.amazon.awssdk.core.SdkRequest;
 import software.amazon.awssdk.core.SdkResponse;
 import software.amazon.awssdk.core.interceptor.*;
 
 public class AdotAwsSdkTracingExecutionInterceptor implements ExecutionInterceptor {
 
+  private static final String GEN_AI_SYSTEM_BEDROCK = "aws.bedrock";
+  private static final ExecutionAttribute<AwsSdkRequest> AWS_SDK_REQUEST_ATTRIBUTE =
+      new ExecutionAttribute<>(
+          AdotAwsSdkTracingExecutionInterceptor.class.getName() + ".AwsSdkRequest");
+
+  private final FieldMapper fieldMapper = new FieldMapper();
+
   // This is the latest point we can obtain the Sdk Request after it is modified by the upstream
   // TracingInterceptor. It ensures upstream handles the request and applies its changes first.
   @Override
   public void beforeTransmission(
-      Context.BeforeTransmission context, ExecutionAttributes executionAttributes) {}
+      Context.BeforeTransmission context, ExecutionAttributes executionAttributes) {
+
+    SdkRequest request = context.request();
+    Span currentSpan = Span.current();
+
+    try {
+      // Skip injection if Otel span is invalid
+      if (currentSpan != null && currentSpan.getSpanContext().isValid()) {
+        AwsSdkRequest awsSdkRequest = AwsSdkRequest.ofSdkRequest(request);
+        if (awsSdkRequest != null) {
+          executionAttributes.putAttribute(AWS_SDK_REQUEST_ATTRIBUTE, awsSdkRequest);
+          fieldMapper.mapToAttributes(request, awsSdkRequest, currentSpan);
+          if (awsSdkRequest.type() == BEDROCKRUNTIME) {
+            currentSpan.setAttribute(GEN_AI_SYSTEM, GEN_AI_SYSTEM_BEDROCK);
+          }
+        }
+      }
+    } catch (Throwable throwable) {
+      // ignore
+    }
+  }
 
   // This is the latest point we can obtain the Sdk Response before span completion in upstream's
   // afterExecution. This ensures we capture attributes from the final, fully modified response
@@ -32,6 +64,14 @@ public class AdotAwsSdkTracingExecutionInterceptor implements ExecutionIntercept
   @Override
   public SdkResponse modifyResponse(
       Context.ModifyResponse context, ExecutionAttributes executionAttributes) {
+
+    Span currentSpan = Span.current();
+    AwsSdkRequest sdkRequest = executionAttributes.getAttribute(AWS_SDK_REQUEST_ATTRIBUTE);
+
+    if (sdkRequest != null) {
+      fieldMapper.mapToAttributes(context.response(), sdkRequest, currentSpan);
+      executionAttributes.putAttribute(AWS_SDK_REQUEST_ATTRIBUTE, null);
+    }
 
     return context.response();
   }
--- a/instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsExperimentalAttributes.java
+++ b/instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsExperimentalAttributes.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2;
+
+import static io.opentelemetry.api.common.AttributeKey.stringKey;
+
+import io.opentelemetry.api.common.AttributeKey;
+
+final class AwsExperimentalAttributes {
+  static final AttributeKey<String> AWS_BUCKET_NAME = stringKey("aws.bucket.name");
+  static final AttributeKey<String> AWS_QUEUE_URL = stringKey("aws.queue.url");
+  static final AttributeKey<String> AWS_QUEUE_NAME = stringKey("aws.queue.name");
+  static final AttributeKey<String> AWS_STREAM_NAME = stringKey("aws.stream.name");
+  static final AttributeKey<String> AWS_TABLE_NAME = stringKey("aws.table.name");
+  static final AttributeKey<String> AWS_GUARDRAIL_ID = stringKey("aws.bedrock.guardrail.id");
+  static final AttributeKey<String> AWS_GUARDRAIL_ARN = stringKey("aws.bedrock.guardrail.arn");
+  static final AttributeKey<String> AWS_AGENT_ID = stringKey("aws.bedrock.agent.id");
+  static final AttributeKey<String> AWS_DATA_SOURCE_ID = stringKey("aws.bedrock.data_source.id");
+  static final AttributeKey<String> AWS_KNOWLEDGE_BASE_ID =
+      stringKey("aws.bedrock.knowledge_base.id");
+
+  // TODO: Merge in gen_ai attributes in opentelemetry-semconv-incubating once upgrade to v1.26.0
+  static final AttributeKey<String> GEN_AI_MODEL = stringKey("gen_ai.request.model");
+  static final AttributeKey<String> GEN_AI_SYSTEM = stringKey("gen_ai.system");
+
+  static final AttributeKey<String> GEN_AI_REQUEST_MAX_TOKENS =
+      stringKey("gen_ai.request.max_tokens");
+
+  static final AttributeKey<String> GEN_AI_REQUEST_TEMPERATURE =
+      stringKey("gen_ai.request.temperature");
+
+  static final AttributeKey<String> GEN_AI_REQUEST_TOP_P = stringKey("gen_ai.request.top_p");
+
+  static final AttributeKey<String> GEN_AI_RESPONSE_FINISH_REASONS =
+      stringKey("gen_ai.response.finish_reasons");
+
+  static final AttributeKey<String> GEN_AI_USAGE_INPUT_TOKENS =
+      stringKey("gen_ai.usage.input_tokens");
+
+  static final AttributeKey<String> GEN_AI_USAGE_OUTPUT_TOKENS =
+      stringKey("gen_ai.usage.output_tokens");
+
+  static final AttributeKey<String> AWS_STATE_MACHINE_ARN =
+      stringKey("aws.stepfunctions.state_machine.arn");
+
+  static final AttributeKey<String> AWS_STEP_FUNCTIONS_ACTIVITY_ARN =
+      stringKey("aws.stepfunctions.activity.arn");
+
+  static final AttributeKey<String> AWS_SNS_TOPIC_ARN = stringKey("aws.sns.topic.arn");
+
+  static final AttributeKey<String> AWS_SECRET_ARN = stringKey("aws.secretsmanager.secret.arn");
+
+  static final AttributeKey<String> AWS_LAMBDA_NAME = stringKey("aws.lambda.function.name");
+
+  static final AttributeKey<String> AWS_LAMBDA_ARN = stringKey("aws.lambda.function.arn");
+
+  static final AttributeKey<String> AWS_LAMBDA_RESOURCE_ID =
+      stringKey("aws.lambda.resource_mapping.id");
+
+  static boolean isGenAiAttribute(String attributeKey) {
+    return attributeKey.equals(GEN_AI_REQUEST_MAX_TOKENS.getKey())
+        || attributeKey.equals(GEN_AI_REQUEST_TEMPERATURE.getKey())
+        || attributeKey.equals(GEN_AI_REQUEST_TOP_P.getKey())
+        || attributeKey.equals(GEN_AI_RESPONSE_FINISH_REASONS.getKey())
+        || attributeKey.equals(GEN_AI_USAGE_INPUT_TOKENS.getKey())
+        || attributeKey.equals(GEN_AI_USAGE_OUTPUT_TOKENS.getKey());
+  }
+
+  private AwsExperimentalAttributes() {}
+}
--- a/instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsJsonProtocolFactoryAccess.java
+++ b/instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsJsonProtocolFactoryAccess.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2;
+
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Source: https://github.com/open-telemetry/opentelemetry-java-instrumentation
+ */
+
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import java.net.URI;
+import javax.annotation.Nullable;
+import software.amazon.awssdk.core.client.config.SdkClientConfiguration;
+import software.amazon.awssdk.core.client.config.SdkClientOption;
+import software.amazon.awssdk.http.SdkHttpFullRequest;
+import software.amazon.awssdk.http.SdkHttpMethod;
+import software.amazon.awssdk.protocols.core.OperationInfo;
+import software.amazon.awssdk.protocols.core.ProtocolMarshaller;
+
+final class AwsJsonProtocolFactoryAccess {
+
+  private static final OperationInfo OPERATION_INFO =
+      OperationInfo.builder().hasPayloadMembers(true).httpMethod(SdkHttpMethod.POST).build();
+
+  @Nullable private static final MethodHandle INVOKE_CREATE_PROTOCOL_MARSHALLER;
+
+  static {
+    MethodHandle invokeCreateProtocolMarshaller = null;
+    try {
+      Class<?> awsJsonProtocolFactoryClass =
+          Class.forName("software.amazon.awssdk.protocols.json.AwsJsonProtocolFactory");
+      Object awsJsonProtocolFactoryBuilder =
+          awsJsonProtocolFactoryClass.getMethod("builder").invoke(null);
+      awsJsonProtocolFactoryBuilder
+          .getClass()
+          .getMethod("clientConfiguration", SdkClientConfiguration.class)
+          .invoke(
+              awsJsonProtocolFactoryBuilder,
+              SdkClientConfiguration.builder()
+                  // AwsJsonProtocolFactory requires any URI to be present
+                  .option(SdkClientOption.ENDPOINT, URI.create("http://empty"))
+                  .build());
+      @SuppressWarnings("rawtypes")
+      Class awsJsonProtocolClass =
+          Class.forName("software.amazon.awssdk.protocols.json.AwsJsonProtocol");
+      @SuppressWarnings("unchecked")
+      Object awsJsonProtocol = Enum.valueOf(awsJsonProtocolClass, "AWS_JSON");
+      awsJsonProtocolFactoryBuilder
+          .getClass()
+          .getMethod("protocol", awsJsonProtocolClass)
+          .invoke(awsJsonProtocolFactoryBuilder, awsJsonProtocol);
+      Object awsJsonProtocolFactory =
+          awsJsonProtocolFactoryBuilder
+              .getClass()
+              .getMethod("build")
+              .invoke(awsJsonProtocolFactoryBuilder);
+
+      MethodHandle createProtocolMarshaller =
+          MethodHandles.publicLookup()
+              .findVirtual(
+                  awsJsonProtocolFactoryClass,
+                  "createProtocolMarshaller",
+                  MethodType.methodType(ProtocolMarshaller.class, OperationInfo.class));
+      invokeCreateProtocolMarshaller =
+          createProtocolMarshaller.bindTo(awsJsonProtocolFactory).bindTo(OPERATION_INFO);
+    } catch (Throwable t) {
+      // Ignore;
+    }
+    INVOKE_CREATE_PROTOCOL_MARSHALLER = invokeCreateProtocolMarshaller;
+  }
+
+  @SuppressWarnings("unchecked")
+  @Nullable
+  static ProtocolMarshaller<SdkHttpFullRequest> createMarshaller() {
+    if (INVOKE_CREATE_PROTOCOL_MARSHALLER == null) {
+      return null;
+    }
+
+    try {
+      return (ProtocolMarshaller<SdkHttpFullRequest>) INVOKE_CREATE_PROTOCOL_MARSHALLER.invoke();
+    } catch (Throwable t) {
+      return null;
+    }
+  }
+
+  private AwsJsonProtocolFactoryAccess() {}
+}
--- a/instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsSdkRequest.java
+++ b/instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsSdkRequest.java
@@ -1,0 +1,135 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2;
+
+import static software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsSdkRequestType.BEDROCK;
+import static software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsSdkRequestType.BEDROCKAGENTOPERATION;
+import static software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsSdkRequestType.BEDROCKAGENTRUNTIMEOPERATION;
+import static software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsSdkRequestType.BEDROCKDATASOURCEOPERATION;
+import static software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsSdkRequestType.BEDROCKKNOWLEDGEBASEOPERATION;
+import static software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsSdkRequestType.BEDROCKRUNTIME;
+import static software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsSdkRequestType.DYNAMODB;
+import static software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsSdkRequestType.KINESIS;
+import static software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsSdkRequestType.LAMBDA;
+import static software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsSdkRequestType.S3;
+import static software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsSdkRequestType.SECRETSMANAGER;
+import static software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsSdkRequestType.SNS;
+import static software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsSdkRequestType.SQS;
+import static software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AwsSdkRequestType.STEPFUNCTION;
+import static software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.FieldMapping.request;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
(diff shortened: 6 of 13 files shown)


instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
```diff
@@ -15,16 +15,48 @@
...
 
+  private static final String GEN_AI_SYSTEM_BEDROCK = "aws.bedrock";
+  private static final ExecutionAttribute<AwsSdkRequest> AWS_SDK_REQUEST_ATTRIBUTE =
+      new ExecutionAttribute<>(
+          AdotAwsSdkTracingExecutionInterceptor.class.getName() + ".AwsSdkRequest");
+
+  private final FieldMapper fieldMapper = new FieldMapper();
+
   // This is the latest point we can obtain the Sdk Request after it is modified by the upstream
   // TracingInterceptor. It ensures upstream handles the request and applies its changes first.
```
pxaws: Could you add a reference to different hooks provided by AWS SDK for the lifecycle of a AWS request? And also point to the code in the upstream where they modified the request? I think this will make it easier for later maintenance.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
```diff
@@ -15,16 +15,48 @@
...
   @Override
   public void beforeTransmission(
-      Context.BeforeTransmission context, ExecutionAttributes executionAttributes) {}
+      Context.BeforeTransmission context, ExecutionAttributes executionAttributes) {
+
+    SdkRequest request = context.request();
+    Span currentSpan = Span.current();
+
+    try {
+      // Skip injection if Otel span is invalid
```
pxaws: An edge case here: what if customers choose to disable the upstream aws sdk instrumentation? The span here will not be an upstream aws sdk span. Is it possible to add a condition check for aws sdk span?

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
```diff
@@ -15,16 +15,48 @@
...
+import io.opentelemetry.api.trace.Span;
+import software.amazon.awssdk.core.SdkRequest;
 import software.amazon.awssdk.core.SdkResponse;
 import software.amazon.awssdk.core.interceptor.*;
 
 public class AdotAwsSdkTracingExecutionInterceptor implements ExecutionInterceptor {
 
+  private static final String GEN_AI_SYSTEM_BEDROCK = "aws.bedrock";
+  private static final ExecutionAttribute<AwsSdkRequest> AWS_SDK_REQUEST_ATTRIBUTE =
+      new ExecutionAttribute<>(
```
pxaws, line 65: Could you add some comments to describe what do ExecutionAttribute and FieldMapper do? Or maybe in some readme file (if you have plan to document how the whole patching works)?

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
```diff
@@ -33,6 +65,14 @@ public void beforeTransmission(
   public SdkResponse modifyResponse(
       Context.ModifyResponse context, ExecutionAttributes executionAttributes) {
 
+    Span currentSpan = Span.current();
+    AwsSdkRequest sdkRequest = executionAttributes.getAttribute(AWS_SDK_REQUEST_ATTRIBUTE);
+
+    if (sdkRequest != null) {
+      fieldMapper.mapToAttributes(context.response(), sdkRequest, currentSpan);
```
pxaws: Just curious: why we call `fieldMapper.mapToAttributes` twice? One inside `beforeTransmission` and the other one in `modifyResponse`.

anahatAWS: Similar to what upstream does [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.11.x/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/internal/TracingExecutionInterceptor.java#L382). If the request attribute is not null, then it uses field mapper to inject the response attribute into the span. 

It helps us extract and map the response attributes.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
anahatAWS: I will the documentation for the hooks and the references to the last points of upstream's sdk request/response modifications to the README as well.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
anahatAWS: Yes, I have added the condition where the it checks the executionAttributes to see if upstream aws-sdk instrumentation is enabled. If the execution attributes do no contain the upstream specific attribute, I think it is safe to say that span is not an aws sdk span.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
anahatAWS: Added in README

instrumentation/aws-sdk/README.md
```diff
@@ -31,8 +31,13 @@ The AdotAwsSdkTracingExecutionInterceptor hooks onto OpenTelemetry's spans durin
...
+All the span lifecycle hooks provided by AWS SDK ExecutionInterceptor can be found [here.](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/core/interceptor/ExecutionInterceptor.html)
 
+_**Important Notes:**_
+- The upstream interceptor's last point of request modification occurs in [beforeTransmission](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.11.x/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/internal/TracingExecutionInterceptor.java#L237).
+- The upstream interceptor closes the span in [afterExecution](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.11.x/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/internal/TracingExecutionInterceptor.java#L348). That hook is inaccessible for span modification.
+`modifyResponse` is our final hook point, giving us access to both the fully processed response and active span.
 
+Variables in AdotAwsSdkTracingExecutionInterceptor:
```
pxaws: I am not sure if this contains enough details for people to understand. Is it possible to draw a high-level sequence diagram to illustrate the how the classes `FieldMapper`, `FieldMapping`, `MethodHandleFactory`, ... work together to set the span attribute for an aws sdk request? You don't need to specify the low-level details about which methods are called. But it would be very helpful to illustrate the main functionalities of each classes and their relationships (e.g. who calls/relies on who for what?)

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
```diff
@@ -15,24 +15,104 @@
...
+
+    try {
+      if (currentSpan == null || !currentSpan.getSpanContext().isValid()) {
+        return;
+      }
+
+      // Check if upstream aws sdk instrumentation is enabled
+      if (!executionAttributes
+          .toString()
+          .contains("io.opentelemetry.javaagent.shaded.instrumentation.awssdk")) {
```
pxaws: Could you add comment to explain why this attribute is expected for a span created by upstream instrumentation? `io.opentelemetry.javaagent.shaded.instrumentation.awssdk` looks like a shaded class name. Will the name be stable in future?

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
```diff
@@ -15,24 +15,104 @@
...
+      Region signingRegion =
+          executionAttributes.getAttribute(AwsSignerExecutionAttribute.SIGNING_REGION);
 
-  // This is the latest point we can obtain the Sdk Response before span completion in upstream's
-  // afterExecution. This ensures we capture attributes from the final, fully modified response
-  // after all upstream interceptors have processed it.
+      if (credentials != null) {
+        String accessKeyId = credentials.accessKeyId();
+        if (accessKeyId != null) {
+          currentSpan.setAttribute(AWS_AUTH_ACCESS_KEY, accessKeyId);
```
pxaws: Is this the new change required by cross-account support?

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsJsonProtocolFactoryAccess.java
pxaws, line 1: Though this class is copied from upstream, I think it's better to have a high-level description about the functionalities of this class. In the normal case, we shouldn't care about the implementation details. But I would like people to understand the big picture: for example, it seems to me that class uses reflection to access some internal method in aws sdk so that it can access the marshalled json response? This is useful because we will need to access those bedrock response to set span attributes?

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/MethodHandleFactory.java
pxaws, line 1: Better to add some high-level description for this class

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/Serializer.java
pxaws, line 1: It seems like the bedrock serializer is quite different from the serializer from upstream. Can we make it a separate class instead of modifying the upstream class?

pxaws: In addition, in the fieldMapper, we have the following code: 

```
      if (AwsExperimentalAttributes.isGenAiAttribute(fieldMapping.getAttribute())) {
         value = serializer.serialize(fieldMapping.getAttribute(), target);
      } else {
         value = serializer.serialize(target);
      }
```

If we separate out the serializer for bedrock, do we need need to call the upstream code `value = serializer.serialize(target);` because it is supposed to be called in upstream instrumentation already?

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsSdkRequestType.java
```diff
@@ -0,0 +1,126 @@
...
+import static software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.FieldMapping.request;
+import static software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.FieldMapping.response;
+
+import io.opentelemetry.api.common.AttributeKey;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+enum AwsSdkRequestType {
+  S3(request(AWS_BUCKET_NAME.getKey(), "Bucket")),
```
pxaws, line 61: Seems like upstream also set the same attribute: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.11.x/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/internal/AwsSdkRequestType.java#L16. Do we have to keep it here?

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
anahatAWS: Yes they are. They were added to the upstream aws-sdk-2.2 TracingExecutionInterceptor in the patch [here](https://github.com/anahatAWS/aws-otel-java-instrumentation/blob/spi-2.2-patch/.github/patches/opentelemetry-java-instrumentation.patch#L3440)

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsSdkRequestType.java
anahatAWS: Addressed offline, these are needed to access the new, non-bedrock experimental attributes

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/Serializer.java
anahatAWS: Addressed offline, the general serialization (serialize(target)) should be kept alongside the GenAI-specific serialization

instrumentation/aws-sdk/README.md
anahatAWS: I have updated the README that includes a sequence diagram and a high-level explanation of the functionality of each class.

instrumentation/aws-sdk/sequence-diagram-2.2.png
pxaws, line 1: Could you add some high-level description for the arrows in the sequence diagram?

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
pxaws: Any alternative way to check? It seems not very stable to rely on the shaded class name. It's quite easy for upstream to just use another name without knowing we are relying on the old name. The whole shading things should be transparent to business logic.

anahatAWS: addressed offline, will check the span attributes for aws-specific attributes that should be set by upstream aws-sdk. Span interface is designed for setting attributes, not reading them ([Span API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span)). It would have been more efficient if the api allowed us to search the map, but I think the approach I suggested is a way around it

anahatAWS: Addressed after discussion: we will be adding a note in the README instead of doing this check (the check could add more unneeded overhead for ADOT if spans are large). This is very rare edge case, so we are accepting the trade-off of leaving a note instead of hard coding the check.

pxaws (approved):

instrumentation/aws-sdk/README.md
```diff
@@ -15,6 +15,10 @@ _Initialization Workflow_
 
 ### AWS SDK v2 Instrumentation Summary
 
+_IMPORTANT:_ This instrumentation depends on upstream OpenTelemetry AWS SDK instrumentation
+being enabled. However, if you disable upstream aws-sdk spans, you must also disable this instrumentation. Even with upstream disabled, this interceptor
+will still be loaded by AWS SDK's SPI mechanism and attempt to enhance spans with AWS attributes.
+
```
thpierce: This doesn't really make a lot of sense. Do we instrument spans other than those created by aws-sdk?

instrumentation/aws-sdk/build.gradle.kts
```diff
@@ -16,12 +16,19 @@
...
   id("com.gradleup.shadow")
+  id("groovy")
 }
 
 base.archivesBaseName = "aws-instrumentation-aws-sdk"
 
 dependencies {
   compileOnly("io.opentelemetry.javaagent:opentelemetry-javaagent-extension-api")
   compileOnly("software.amazon.awssdk:aws-core:2.2.0")
+  compileOnly("software.amazon.awssdk:secretsmanager:2.2.0")
```
thpierce: Why is secretsmanager needed and not any other?

instrumentation/aws-sdk/build.gradle.kts
```diff
@@ -16,12 +16,19 @@
...
+
   compileOnly("net.bytebuddy:byte-buddy")
+  compileOnly("com.google.code.findbugs:jsr305:3.0.2")
```
thpierce, line 32: Why is this needed?

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkInstrumentationModule.java
```diff
@@ -37,13 +37,29 @@ public AdotAwsSdkInstrumentationModule() {
...
-    return 99;
+    return Integer.MAX_VALUE;
   }
 
   @Override
   public List<String> getAdditionalHelperClassNames() {
     return Arrays.asList(
-        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AdotAwsSdkTracingExecutionInterceptor");
+        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AdotAwsSdkTracingExecutionInterceptor",
+        "software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2.AdotTracingExecutionInterceptor$RequestSpanFinisher",
```
thpierce: What is AdotTracingExecutionInterceptor$RequestSpanFinisher?

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsExperimentalAttributes.java
thpierce, line 1: Confirmed exact match to patch.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsJsonProtocolFactoryAccess.java
thpierce, line 1: Confirmed exact match with upstream.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsSdkRequest.java
```diff
@@ -0,0 +1,135 @@
...
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
```
thpierce, line 14: Need to also attribute upstream. Applies to other classes, too.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsSdkRequestType.java
thpierce, line 1: Missing https://github.com/aws-observability/aws-otel-java-instrumentation/blob/ac3c0c75a977342e58af137973e64fa049e94f1e/.github/patches/opentelemetry-java-instrumentation.patch#L2849C1-L2849C61 - please keep files identical.

thpierce, line 1: Missing https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/c3bfa1c32d581ede4fb84e504579362d3c8115cb/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/internal/AwsSdkRequestType.java#L27  - please keep files identical.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsSdkRequest.java
thpierce, line 1: Missing https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/c3bfa1c32d581ede4fb84e504579362d3c8115cb/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/internal/AwsSdkRequest.java#L37-L122 and maybe more - please keep files identical.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/BedrockJsonParser.java
thpierce, line 1: Confirmed exact match to patch.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/FieldMapper.java
thpierce, line 1: Confirmed exact match to upstream + patch.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/FieldMapping.java
thpierce, line 1: Confirmed exact match to upstream.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/MethodHandleFactory.java
thpierce, line 1: Confirmed exact match to upstream.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/Serializer.java
thpierce, line 1: Confirmed exact match to upstream + patch.

instrumentation/aws-sdk/src/test/groovy/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/BedrockJsonParserTest.groovy
```diff
@@ -0,0 +1,107 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
```
thpierce: Does not match other attributions.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
```diff
@@ -15,24 +15,97 @@
...
   public SdkResponse modifyResponse(
       Context.ModifyResponse context, ExecutionAttributes executionAttributes) {
 
+    Span currentSpan = Span.current();
+    AwsSdkRequest sdkRequest = executionAttributes.getAttribute(AWS_SDK_REQUEST_ATTRIBUTE);
+
+    if (sdkRequest != null) {
+      fieldMapper.mapToAttributes(context.response(), sdkRequest, currentSpan);
+      executionAttributes.putAttribute(AWS_SDK_REQUEST_ATTRIBUTE, null);
+    }
```
thpierce: I don't get why this is needed. Update method comment to explain why this is needed/correct. Same goes with beforeTransmission

instrumentation/aws-sdk/build.gradle.kts
anahatAWS: it is needed for `import javax.annotation.Nullable;` to use @Nullable  in FieldMapper, AwsSdkRequest, Serializer, and AwsJsonProtocolFactoryAccess,

instrumentation/aws-sdk/README.md
anahatAWS: Addressed offline: will remove. It wouldn't make sense as our instrumentation wouldn't inject anything because there would be no aws-sdk spans.

instrumentation/aws-sdk/build.gradle.kts
anahatAWS: I had all the others included, but I wanted to check if each one was needed and secretsmanager seemed to be the only one required. But after further investigation, I see why it fails to build without this dependency. We need the `compileOnly("software.amazon.awssdk:aws-json-protocol:2.2.0")` instead for classes related to AWS JSON protocol handling. The reason it worked with secretsmanager was because secretsmanager includes aws-json-protocol as a dependency [here](https://central.sonatype.com/artifact/software.amazon.awssdk/secretsmanager/2.2.0/dependencies).

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkInstrumentationModule.java
anahatAWS: It was an inner class I previously had. It has been removed now, I will also remove it from here.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsSdkRequest.java
anahatAWS: Those were intentionally removed because upstream will be handling those requests. The requests in ADOT are specific to the new experimental attributes that were added. This was done to avoid duplicate work.

instrumentation/aws-sdk/build.gradle.kts
thpierce: Makes more sense. Ty

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsSdkRequest.java
thpierce: Oh got it, yes that makes sense.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsSdkRequest.java
```diff
@@ -0,0 +1,144 @@
...
+
+package software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2;
+
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Modifications Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ */
+
```
thpierce, line 24: Should this not be like this (applies to many files):

```suggestion
/*
 * Copyright The OpenTelemetry Authors
 * SPDX-License-Identifier: Apache-2.0
 *
 * Modifications Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 */
 
package software.amazon.opentelemetry.javaagent.instrumentation.awssdk_v2_2;
```

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsSdkRequest.java
```diff
@@ -0,0 +1,144 @@
...
+import software.amazon.awssdk.core.SdkRequest;
+
+@SuppressWarnings("MemberName")
+enum AwsSdkRequest {
+  // generic requests
+  DynamoDbRequest(DYNAMODB, "DynamoDbRequest"),
+  S3Request(S3, "S3Request"),
+  SnsRequest(SNS, "SnsRequest"),
+  SqsRequest(SQS, "SqsRequest"),
+  KinesisRequest(KINESIS, "KinesisRequest"),
```
thpierce, line 54: Why do we keep `generic requests`, but not `specific requests`?

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
```diff
@@ -15,24 +15,102 @@
...
+   *
+   * <p>Upstream's last Sdk Request modification: @see <a
+   * href="https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.11.x/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/internal/TracingExecutionInterceptor.java#L237">reference</a>
+   */
   @Override
   public void beforeTransmission(
-      Context.BeforeTransmission context, ExecutionAttributes executionAttributes) {}
+      Context.BeforeTransmission context, ExecutionAttributes executionAttributes) {
+
+    SdkRequest request = context.request();
```
thpierce: Should we null check?

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
```diff
@@ -15,24 +15,102 @@
...
+    Span currentSpan = Span.current();
+
+    try {
+      if (currentSpan == null || !currentSpan.getSpanContext().isValid()) {
+        return;
```
thpierce: What is this scenario?

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
```diff
@@ -15,24 +15,102 @@
...
+      }
+
+      AwsCredentials credentials =
```
thpierce: Should this only be run if `if (captureExperimentalSpanAttributes) {`?

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
```diff
@@ -15,24 +15,102 @@
...
+   * <p>Upstream's last Sdk Request modification: @see <a
+   * href="https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.11.x/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/internal/TracingExecutionInterceptor.java#L237">reference</a>
+   */
   @Override
   public void beforeTransmission(
-      Context.BeforeTransmission context, ExecutionAttributes executionAttributes) {}
+      Context.BeforeTransmission context, ExecutionAttributes executionAttributes) {
+
+    SdkRequest request = context.request();
+    Span currentSpan = Span.current();
```
thpierce: Why is this equivalent to `Span span = Span.fromContext(getContext(executionAttributes));`

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
thpierce: Should this be `context.httpRequest();`?

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
```diff
@@ -15,24 +15,102 @@
...
+   *
+   * <p>Upstream's last Sdk Response modification before span closure: @see <a
+   * href="https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.11.x/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/internal/TracingExecutionInterceptor.java#L348">reference</a>
+   */
   @Override
   public SdkResponse modifyResponse(
       Context.ModifyResponse context, ExecutionAttributes executionAttributes) {
 
+    Span currentSpan = Span.current();
+    AwsSdkRequest sdkRequest = executionAttributes.getAttribute(AWS_SDK_REQUEST_ATTRIBUTE);
```
thpierce: Ok so in beforeTransmission() we are storing `AWS_SDK_REQUEST_ATTRIBUTE` (which is the entire request) in the span, the using the request to add other attributes to the span. Then in `modifyResponse` we are pulling it out, using it + the response to add more attributes to the span, and then removing `AWS_SDK_REQUEST_ATTRIBUTE`, since it's not needed. Is that roughly correct?

What I'm mostly confused by is: Does the upstream NOT do this for us? I see similar logic already, but maybe it's added and removed before our interceptors take action? I also see logic to remove this attribute, when does that run? https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.11.x/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/internal/TracingExecutionInterceptor.java#L440

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsSdkRequest.java
anahatAWS: Addressed offline: backlog edit

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
anahatAWS: This is a safe guard check

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
anahatAWS: Different way of handling than compared to upstream since we don't have access to otel context. We want to access the current spans to inject into.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
anahatAWS: Addressed offline: different handling than upstream done on purpose

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
anahatAWS: Addressed offline: will rename the attribute to eliminate confusion between upstream and adot

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsSdkRequest.java
anahatAWS: Upstream handles the other specific requests. These generic requests were kept because ADOT adds new attributes for them.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
thpierce: Much more clear, ty.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AdotAwsSdkTracingExecutionInterceptor.java
```diff
@@ -51,41 +83,47 @@ public class AdotAwsSdkTracingExecutionInterceptor implements ExecutionIntercept
...
-        return;
-      }
+      try {
+        if (request == null) {
+          return;
+        }
+
+        if (currentSpan == null || !currentSpan.getSpanContext().isValid()) {
+          return;
+        }
```
thpierce: Not blocking:
```suggestion
        if (request == null || currentSpan == null || !currentSpan.getSpanContext().isValid()) {
          return;
        }
```

thpierce (dismissed):

PR #1115, opened 2025-07-14
Base of AWS SDK v1.11 SPI Implementation

anahatAWS:
This PR is similar to #1111, as it sets a base SPI implementation for AWS SDK v1.11.

### Issue
This is the skeleton set up for the SPI, which aims to remove the [patch](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/patches/opentelemetry-java-instrumentation.patch) for aws-sdk v1.11 by using OTel's InstrumentationModule SPI extension. This instrumentation is essentially complementing the upstream java agent. It is completely separate from upstream and instruments after the Otel agent.

### Description of Changes
This PR sets up the foundational structure for AWS SDK v1.11 Instrumentation, moving away from the current patching approach. It doesn't modify current ADOT functionality or the upstream span. It just registers the ADOT SPI  implementation and sets up the interceptor hooks.

#### Core Components

1. **AdotAwsSdkInstrumentationModule**
   - Extends OpenTelemetry's `InstrumentationModule` SPI
   - Registers custom handler through AdotAwsSdkClientInstrumentation class in `typeInstrumentations` method

2. **AdotAwsSdkClientInstrumentation**
    - AdotAwsSdkClientAdvice registers our handler only if the upstream aws-sdk span is enabled (i.e. it checks if the upstream handler is present when an AWS SDK client is initialized).

3. **AdotAwsSdkTracingRequestHandler**
   - Extends AWS SDK's `RequestHandler2`
   - Hooks into key SDK lifecycle points:
     - `beforeRequest`: Captures final SDK request after upstream modifications
     - `afterAttempt`: Processes response before span closure in upstream
   - Will be used to enriches spans
   - Acts as central coordinator for all the awssdk_v1_11 components

4. **Resources Folder**
    -  Registers the v1.11 AdotAwsSdkInstrumentationModule into OTel's SPI extension classpath in META-INF/services

### Key Design Decisions
1. **Instrumentation Ordering**
   - Deliberately structured to run after upstream OTel agent
   - Ensures all upstream modifications are captured
   - Maintains compatibility with existing instrumentation

2. **Lifecycle Hook Points**
   - `beforeRequest`: Last point to access modified request
   - `afterAttempt`: Final opportunity to enrich span before closure

### Testing
- Verified existing functionality remains unchanged and contract tests pass (all contract tests pass after following the steps [here](https://github.com/aws-observability/aws-otel-java-instrumentation/tree/main/appsignals-tests))
- Confirmed build success with new structure

### DIFF
--- a/instrumentation/aws-sdk/README.md
+++ b/instrumentation/aws-sdk/README.md
@@ -3,7 +3,7 @@
 ### Overview
 The aws-sdk instrumentation is an SPI-based implementation that extends the upstream OpenTelemetry AWS Java SDK instrumentation.
 
-_Initialization Workflow_
+_Example: v2.2 Initialization Workflow_
 
 1. OpenTelemetry Agent starts
    - Loads default instrumentations
@@ -12,6 +12,30 @@ _Initialization Workflow_
 2. Scans for other SPI implementations
    - Finds ADOT’s **AdotAwsSdkInstrumentationModule**
    - Registers **AdotAwsSdkTracingExecutionInterceptor** (order > 0)
+
+### AWS SDK v1 Instrumentation Summary
+The AdotAwsSdkInstrumentationModule uses the instrumentation (specified in AdotAwsClientInstrumentation) to register the AdotAwsSdkTracingRequestHandler through `typeInstrumentations`.
+
+Key aspects of handler registration:
+- `order` method ensures ADOT instrumentation runs after OpenTelemetry's base instrumentation. It is set to 99 as precaution, in case upstream aws-sdk registers more handlers.
+- `AdotAwsSdkClientInstrumentation` class
+  - `AdotAwsSdkClientAdvice` registers our handler only if the upstream aws-sdk span is enabled (i.e. it checks if the upstream handler is present when an AWS SDK client is
+  initialized).
+  - Ensures the OpenTelemetry handler is registered first.
+
+**AdotAwsSdkTracingRequestHandler**
+
+The AdotAwsSdkTracingRequestHandler hooks onto OpenTelemetry's spans during specific phases of the SDK request and response life cycle. These hooks are strategically chosen to ensure proper ordering of attribute injection.
+
+1.  `beforeRequest`: the latest point where the SDK request can be obtained after it is modified by the upstream aws-sdk v1.11 handler
+2.  `afterAttempt`: the latest point to access the SDK response before the span closes in the upstream afterResponse/afterError methods
+
+All the span lifecycle hooks provided by AWS SDK RequestHandler2 can be found [here.](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/handlers/RequestHandler2.html#beforeMarshalling-com.amazonaws.AmazonWebServiceRequest)
+
+_**Important Notes:**_
+- The upstream interceptor's last point of request modification occurs in [beforeRequest](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-sdk/aws-sdk-1.11/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v1_11/TracingRequestHandler.java#L58).
+- The upstream interceptor closes the span in [afterResponse](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-sdk/aws-sdk-1.11/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v1_11/TracingRequestHandler.java#L116) and/or [afterError](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-sdk/aws-sdk-1.11/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v1_11/TracingRequestHandler.java#L131). These hooks are inaccessible for span modification.
+  `afterAttempt` is our final hook point, giving us access to both the fully processed response and active span.
 
 ### AWS SDK v2 Instrumentation Summary
 
--- a/instrumentation/aws-sdk/README.md
+++ b/instrumentation/aws-sdk/README.md
@@ -3,7 +3,21 @@
 ### Overview
 The aws-sdk instrumentation is an SPI-based implementation that extends the upstream OpenTelemetry AWS Java SDK instrumentation.
 
-_Example: v2.2 Initialization Workflow_
+##### _v1.11 Initialization Workflow_
+1. OpenTelemetry Agent Starts 
+    - Loads default instrumentations 
+    - Loads aws-sdk v1.11 instrumentations
+    - Injects **TracingRequestHandler** into constructor
+2. Scans for other SPI implementations
+    - Finds ADOT’s **AdotAwsSdkInstrumentationModule**
+    - Injects code that:
+      - Checks for TracingRequestHandler
+      - If present, adds **AdotAwsSdkTracingRequestHandler**
+3. AWS SDK Client Created 
+   - Constructor runs with injected code:
+      [AWS Handlers] → TracingRequestHandler → AdotAwsSdkTracingRequestHandler
+
+##### _v2.2 Initialization Workflow_
 
 1. OpenTelemetry Agent starts
    - Loads default instrumentations
@@ -17,8 +31,8 @@ _Example: v2.2 Initialization Workflow_
 The AdotAwsSdkInstrumentationModule uses the instrumentation (specified in AdotAwsClientInstrumentation) to register the AdotAwsSdkTracingRequestHandler through `typeInstrumentations`.
 
 Key aspects of handler registration:
-- `order` method ensures ADOT instrumentation runs after OpenTelemetry's base instrumentation. It is set to 99 as precaution, in case upstream aws-sdk registers more handlers.
-- `AdotAwsSdkClientInstrumentation` class
+- `order` method ensures ADOT instrumentation runs after OpenTelemetry's base instrumentation. It is set to the max integer value, as precaution, in case upstream aws-sdk registers more handlers.
+- `AdotAwsSdkClientInstrumentation` class adds ADOT handler to list of request handlers
 
 **AdotAwsSdkClientInstrumentation**
 
@@ -34,6 +48,11 @@ The AdotAwsSdkTracingRequestHandler hooks onto OpenTelemetry's spans during spec
 
 1.  `beforeRequest`: the latest point where the SDK request can be obtained after it is modified by the upstream aws-sdk v1.11 handler
 2.  `afterAttempt`: the latest point to access the SDK response before the span closes in the upstream afterResponse/afterError methods
+    - _NOTE:_ We use afterAttempt not because it's ideal, but because it our last chance to add attributes, even though this means our logic runs multiple times during retries. 
+    - This is a trade-off:
+      - We get to add our attributes before span closure 
+      - But our code runs redundantly on each retry attempt 
+      - We're constrained by when upstream closes the span
 
 All the span lifecycle hooks provided by AWS SDK RequestHandler2 can be found [here.](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/handlers/RequestHandler2.html#beforeMarshalling-com.amazonaws.AmazonWebServiceRequest)
 
--- a/instrumentation/aws-sdk/README.md
+++ b/instrumentation/aws-sdk/README.md
@@ -19,9 +19,14 @@ The AdotAwsSdkInstrumentationModule uses the instrumentation (specified in AdotA
 Key aspects of handler registration:
 - `order` method ensures ADOT instrumentation runs after OpenTelemetry's base instrumentation. It is set to 99 as precaution, in case upstream aws-sdk registers more handlers.
 - `AdotAwsSdkClientInstrumentation` class
+
+**AdotAwsSdkClientInstrumentation**
+
+AWS SDK v1.11 instrumentation requires ByteBuddy because, unlike v2.2, it doesn't provide an SPI for adding request handlers. While v2.2 uses the ExecutionInterceptor interface and Java's ServiceLoader mechanism, v1.11 maintains a direct list of handlers that can't be modified through a public API. Therefore, we use ByteBuddy to modify the AWS client constructor and inject our handler directly into the requestHandler2s list.
+
   - `AdotAwsSdkClientAdvice` registers our handler only if the upstream aws-sdk span is enabled (i.e. it checks if the upstream handler is present when an AWS SDK client is
   initialized).
-  - Ensures the OpenTelemetry handler is registered first.
+    - Ensures the OpenTelemetry handler is registered first.
 
 **AdotAwsSdkTracingRequestHandler**
 
--- a/instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v1_11/AdotAwsSdkClientInstrumentation.java
+++ b/instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v1_11/AdotAwsSdkClientInstrumentation.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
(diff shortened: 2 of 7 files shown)


instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v1_11/AdotAwsSdkClientInstrumentation.java
```diff
@@ -0,0 +1,81 @@
...
+    transformer.applyAdviceToMethod(
+        isConstructor(),
+        AdotAwsSdkClientInstrumentation.class.getName() + "$AdotAwsSdkClientAdvice");
+  }
+
+  @SuppressWarnings("unused")
+  public static class AdotAwsSdkClientAdvice {
+
+    @Advice.OnMethodExit(suppress = Throwable.class)
+    public static void addHandler(
```
pxaws, line 73: Could you explain why we need to use `bytebuddy` to modify the class of aws sdk to add another handler? Can we do the instrumentation non invasively like we do for aws sdk 2.2 (where we rely on Java SPI provided by aws sdk to add the instrumentation hooks)?

anahatAWS: We need ByteBuddy because the aws sdk for v1.11 doesn't doesn't have a built in SPI registration mechanism (like software.amazon.awssdk.global.handlers classpath in resources folder for v2.2). v1.11 actually keeps a list of handlers instead and there is no public API to register handlers globally. I added this spi handler similarly to the upstream handler registration [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v1_11/AwsClientInstrumentation.java). This approach lets us modify the constructor and add our handler consistently, for each client call.

anahatAWS: Updated README to explain this

pxaws: Could you add the upstream code reference as comments in the code?

anahatAWS: Done.

pxaws (dismissed):

instrumentation/aws-sdk/README.md
```diff
@@ -3,7 +3,7 @@
 ### Overview
 The aws-sdk instrumentation is an SPI-based implementation that extends the upstream OpenTelemetry AWS Java SDK instrumentation.
 
-_Initialization Workflow_
+_Example: v2.2 Initialization Workflow_
```
thpierce: Should we add a 1.1 workflow?

instrumentation/aws-sdk/README.md
```diff
@@ -13,6 +13,35 @@ _Initialization Workflow_
    - Finds ADOT’s **AdotAwsSdkInstrumentationModule**
    - Registers **AdotAwsSdkTracingExecutionInterceptor** (order > 0)
 
+### AWS SDK v1 Instrumentation Summary
+The AdotAwsSdkInstrumentationModule uses the instrumentation (specified in AdotAwsClientInstrumentation) to register the AdotAwsSdkTracingRequestHandler through `typeInstrumentations`.
+
+Key aspects of handler registration:
+- `order` method ensures ADOT instrumentation runs after OpenTelemetry's base instrumentation. It is set to 99 as precaution, in case upstream aws-sdk registers more handlers.
+- `AdotAwsSdkClientInstrumentation` class
```
thpierce: Not blocking: Was there supposed to be more information here?

instrumentation/aws-sdk/README.md
```diff
@@ -13,6 +13,35 @@ _Initialization Workflow_
...
+
+**AdotAwsSdkClientInstrumentation**
+
+AWS SDK v1.11 instrumentation requires ByteBuddy because, unlike v2.2, it doesn't provide an SPI for adding request handlers. While v2.2 uses the ExecutionInterceptor interface and Java's ServiceLoader mechanism, v1.11 maintains a direct list of handlers that can't be modified through a public API. Therefore, we use ByteBuddy to modify the AWS client constructor and inject our handler directly into the requestHandler2s list.
```
thpierce, line 39: > AWS SDK v1.11 instrumentation requires ByteBuddy because, unlike v2.2, it doesn't provide an SPI for adding request handlers.

Confused - 2.2 uses ByteBuddy, too.

Also, do we think that upstream SHOULD provide an SPI? Is this something we should contribute to upstream?

instrumentation/aws-sdk/README.md
```diff
@@ -13,6 +13,35 @@ _Initialization Workflow_
    - Finds ADOT’s **AdotAwsSdkInstrumentationModule**
    - Registers **AdotAwsSdkTracingExecutionInterceptor** (order > 0)
 
+### AWS SDK v1 Instrumentation Summary
+The AdotAwsSdkInstrumentationModule uses the instrumentation (specified in AdotAwsClientInstrumentation) to register the AdotAwsSdkTracingRequestHandler through `typeInstrumentations`.
+
+Key aspects of handler registration:
+- `order` method ensures ADOT instrumentation runs after OpenTelemetry's base instrumentation. It is set to 99 as precaution, in case upstream aws-sdk registers more handlers.
```
thpierce: Not blocking: is 99 the max?

instrumentation/aws-sdk/README.md
```diff
@@ -13,6 +13,35 @@ _Initialization Workflow_
...
+  - `AdotAwsSdkClientAdvice` registers our handler only if the upstream aws-sdk span is enabled (i.e. it checks if the upstream handler is present when an AWS SDK client is
+  initialized).
+    - Ensures the OpenTelemetry handler is registered first.
+
+**AdotAwsSdkTracingRequestHandler**
+
+The AdotAwsSdkTracingRequestHandler hooks onto OpenTelemetry's spans during specific phases of the SDK request and response life cycle. These hooks are strategically chosen to ensure proper ordering of attribute injection.
+
+1.  `beforeRequest`: the latest point where the SDK request can be obtained after it is modified by the upstream aws-sdk v1.11 handler
+2.  `afterAttempt`: the latest point to access the SDK response before the span closes in the upstream afterResponse/afterError methods
```
thpierce, line 50: `afterAttempt` runs after each attempt - this could mean re-running logic many times. Is that desirable? Is that something that should be called out?

instrumentation/aws-sdk/README.md
```diff
@@ -13,6 +13,35 @@ _Initialization Workflow_
...
+
+All the span lifecycle hooks provided by AWS SDK RequestHandler2 can be found [here.](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/handlers/RequestHandler2.html#beforeMarshalling-com.amazonaws.AmazonWebServiceRequest)
```
thpierce, line 57: Very helpful, thanks, possible to provide a similar link for V2?

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v1_11/AdotAwsSdkInstrumentationModule.java
```diff
@@ -0,0 +1,54 @@
...
+import static io.opentelemetry.javaagent.extension.matcher.AgentElementMatchers.hasClassesNamed;
+
+import io.opentelemetry.javaagent.extension.instrumentation.InstrumentationModule;
+import io.opentelemetry.javaagent.extension.instrumentation.TypeInstrumentation;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import net.bytebuddy.matcher.ElementMatcher;
+
+public class AdotAwsSdkInstrumentationModule extends InstrumentationModule {
```
thpierce, line 34: Might be good to callout this is adapted from [AbstractAwsSdkInstrumentationModule](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v1_11/AbstractAwsSdkInstrumentationModule.java) (at whatever commit you are basing it off of).

Similar feedback to the other classes. That way if people have questions they can reference the upstream and probably get an answer.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v1_11/AdotAwsSdkClientInstrumentation.java
```diff
@@ -0,0 +1,88 @@
...
+   *
+   * <p>Upstream handler registration: @see
+   * <a href="https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v1_11/AwsClientInstrumentation.java#L39">...</a>
+   */
+  @SuppressWarnings("unused")
+  public static class AdotAwsSdkClientAdvice {
+
+    @Advice.OnMethodExit(suppress = Throwable.class)
+    public static void addHandler(
+        @Advice.FieldValue(value = "requestHandler2s", readOnly = false)
```
thpierce: > , readOnly = false

Why?

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v1_11/AdotAwsSdkClientInstrumentation.java
```diff
@@ -0,0 +1,88 @@
...
+          break;
+        }
+      }
+
+      // Only adds our handler if aws-sdk spans are enabled. This also ensures upstream
+      // instrumentation is applied first.
+      if (hasOtelHandler && !hasAdotHandler) {
+        handlers.add(new AdotAwsSdkTracingRequestHandler());
+      }
+    }
```
thpierce, line 102: Not sure I'm getting this. 

1. What does it meant to enable aws-sdk spans?
2. What is the situation where we already have an ADOT Handler?
3. I'm guessing handlers are an ordered list, so adding ADOT handler only if OTEL handler is present makes a kind of sense, but is there a scenario where we want to add a handler, but there isn't an OTEL handler already?

instrumentation/aws-sdk/README.md
anahatAWS: This order method returns an int value (32-bit integer). So the max is 2,147,483,647. I chose 99 thinking it would be a large enough distance from upstream which uses 0. To be more cautious, I could use Integer.MAX_VALUE if that is preferred.

instrumentation/aws-sdk/README.md
anahatAWS: No, it is meant to highlight the importance of that class, which I dive deeper into below that bullet point.

instrumentation/aws-sdk/README.md
anahatAWS: The SPI not being provided (and having to use ByteBuddy) is regarding the AWS SDK having different interfaces to intercept sdk calls (RequestHandler2 vs ExecutionInterceptor). It is unrelated to core OTel.

In AWS SDK v1 (RequestHandler2), ByteBuddy is required to inject our handler into the SDK's request handler list because there's no public API or SPI to add handlers globally.
 
In AWS SDK v2, ByteBuddy is used by the core OpenTelemetry instrumentation, but not for loading the interceptor. We use the ExecutionInterceptor's  SPI classpath `/software/amazon/awssdk/global/handlers/execution.interceptors` to register out ADOT extension without additional bytecode manipulation.

instrumentation/aws-sdk/README.md
anahatAWS: Yes, it has been added in the second v2.2 PR: https://github.com/aws-observability/aws-otel-java-instrumentation/pull/1113.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v1_11/AdotAwsSdkClientInstrumentation.java
anahatAWS: Initially, after reading the [documentation](https://javadoc.io/doc/net.bytebuddy/byte-buddy/latest/index.html), I added readOnly = false thinking we needed it to modify the requestHandler2s list by adding our handler. However, after investigation, I realize we can use the default readOnly = true because we're only modifying the List's contents through handlers.add(), not reassigning the field itself.

readOnly = false would only be needed if we were trying to reassign the field (e.g., handlers = new ArrayList<>())
But we can modify its contents even when it is =true. So, I believe we can simplify the code by removing the explicit readOnly parameter and using the default value.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v1_11/AdotAwsSdkClientInstrumentation.java
anahatAWS: 1. The check is also an edge case here; if customers choose to disable the upstream aws sdk instrumentation, the span here will not be an upstream aws sdk span. So we want to check the core OpenTelemetry aws sdk instrumentation is active before adding our handler.
2. This prevents duplicate handlers if the constructor is called multiple times. Without this check, we might add multiple ADOT handlers to the same client. It is similar to what upstream did when adding their handler [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v1_11/AwsClientInstrumentation.java#L47)
3. Our ADOT handler enhances the spans created by OTel. I don't see a case or possibility where we wouldn't need the OTel handler, since we are depending on its instrumentation. Without OTel spans to enhance, our handler would have no purpose.

instrumentation/aws-sdk/README.md
anahatAWS: Will do, since v1 and v2 have different workflows

instrumentation/aws-sdk/README.md
anahatAWS: Yes, I will call this out since `afterAttempt` is not ideal, but our only option to add attributes before the span is closed by upstream. The ideal would have been `afterResponse` because it runs only once after all retries are complete. But I will add documentation to this to clarify the trade-off being made. Thanks for pointing it out.

instrumentation/aws-sdk/README.md
thpierce: Also updated in https://github.com/aws-observability/aws-otel-java-instrumentation/pull/1113/files, ty

instrumentation/aws-sdk/README.md
thpierce: Ok I think I get it. Seems reasonable.

thpierce: Discussed offline - based on how AWS SDK V1/V2 are implemented, the insturmentation must be done differently, no upstream improvement reasonable.

instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v1_11/AdotAwsSdkClientInstrumentation.java
```diff
@@ -0,0 +1,94 @@
...
+   * v2.2. AWS SDK v1.11 keeps a list of handlers instead.
+   *
+   * <p>Upstream handler registration: @see <a
+   * href="https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v1_11/AwsClientInstrumentation.java#L39">...</a>
+   */
+  @SuppressWarnings("unused")
+  public static class AdotAwsSdkClientAdvice {
+
+    @Advice.OnMethodExit(suppress = Throwable.class)
+    public static void addHandler(
```
thpierce, line 73: Lets add a few unit tests for this to assert expected behaviour.

thpierce (approved):

PR #1117, opened 2025-07-18
AWS SDK v1.11 Patch Migration

anahatAWS:
Note: This is a continuation of https://github.com/aws-observability/aws-otel-java-instrumentation/pull/1115

### Description of Changes
This implementation builds on the foundation established in PR https://github.com/aws-observability/aws-otel-java-instrumentation/pull/1115, transforming the structural setup into a fully functional SPI-based solution that will replace our current patching approach. This PR does not change the current ADOT functionality because patches have not been removed.

The next/final PR for v1.11 will remove the patches for aws-sdk-1.11 and have unit tests to ensure correct SPI functionality and behaviour. The final PR will also pass all the contract-tests once patches are removed.

#### Changes include:
- Migration of patched files into proper package structure: 
   NOTE: We are not copying entire files from upstream. Instead, we only migrated the new components that were added by our patches and the methods that use these AWS-specific components. I deliberately removed any code that was untouched by our patches to avoid duplicating upstream instrumentation code. This selective migration ensures we maintain only our AWS-specific additions while letting OTel handle its base functionality.
    - `AwsBedrockResourceType` - [patch](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/patches/opentelemetry-java-instrumentation.patch#L224) creates new class
    - `AwsExperimentalAttributes` - [patch](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/patches/opentelemetry-java-instrumentation.patch#L362) on [this otel file](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.11.x/instrumentation/aws-sdk/aws-sdk-1.11/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v1_11/AwsExperimentalAttributes.java)
    - `AwsSdkExperimentalAttributesExtractor` - [patch](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/patches/opentelemetry-java-instrumentation.patch#L408) on [this otel file](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.11.x/instrumentation/aws-sdk/aws-sdk-1.11/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v1_11/AwsSdkExperimentalAttributesExtractor.java)
[...]

PR #1120, opened 2025-07-25
Instrumentation Patch Removal and SPI AWS SDK Test Addition

anahatAWS:
This is the final PR for the SPI aws-sdk instrumentation. It removes the [opentelemetry-java-instrumentation](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/patches/opentelemetry-java-instrumentation.patch) patch and adds comprehensive unit test coverage for AWS experimental attributes in both AWS SDK v1.11 and v2.2 instrumentation packages. The v2.2 package introduces 29 new experimental attributes while v1.11 adds 23 new experimental attributes. All attributes are now tested through unit tests and/or contract tests.

#### AWS SDK v2.2 (awssdk_v2_2)
New attributes being tested:
1. AWS_BUCKET_NAME - testS3ExperimentalAttributes() & contract tests
2. AWS_QUEUE_URL - testSqsExperimentalAttributes() & contract tests
3. AWS_QUEUE_NAME - contract tests
4. AWS_STREAM_NAME - testKinesisExperimentalAttributes() & contract tests
5. AWS_STREAM_ARN - testKinesisExperimentalAttributes() & contract tests
6. AWS_TABLE_NAME - testDynamoDbExperimentalAttributes()
7. AWS_GUARDRAIL_ID - contract tests 
8. AWS_GUARDRAIL_ARN - contract tests
9. AWS_AGENT_ID - testBedrockAgentExperimentalAttributes() & contract tests
10. AWS_DATA_SOURCE_ID - testBedrockDataSourceExperimentalAttributes() & contract tests
11. AWS_KNOWLEDGE_BASE_ID - testBedrockKnowledgeBaseExperimentalAttributes() & contract tests
12. GEN_AI_MODEL - testBedrockExperimentalAttributes() & contract tests
13. GEN_AI_SYSTEM - contract tests
14. GEN_AI_REQUEST_MAX_TOKENS - testBedrockExperimentalAttributes() & contract tests
15. GEN_AI_REQUEST_TEMPERATURE - testBedrockExperimentalAttributes() & contract tests
16. GEN_AI_REQUEST_TOP_P - contract tests
17. GEN_AI_RESPONSE_FINISH_REASONS - contract tests
18. GEN_AI_USAGE_INPUT_TOKENS - contract tests
19. GEN_AI_USAGE_OUTPUT_TOKENS - contract tests
20. AWS_STATE_MACHINE_ARN - testStepFunctionExperimentalAttributes() & contract tests
21. AWS_STEP_FUNCTIONS_ACTIVITY_ARN - testStepFunctionExperimentalAttributes() & contract tests
22. AWS_SNS_TOPIC_ARN - testSnsExperimentalAttributes() & contract tests
23. AWS_SECRET_ARN - testSecretsManagerExperimentalAttributes() & contract tests
24. AWS_LAMBDA_NAME - testLambdaExperimentalAttributes()
25. AWS_LAMBDA_ARN - testLambdaArnExperimentalAttribute()
26. AWS_LAMBDA_RESOURCE_ID - testLambdaResourceIdExperimentalAttribute()
27. AWS_TABLE_ARN - testTableArnExperimentalAttribute()
28. AWS_AUTH_ACCESS_KEY - testAuthAccessKeyExperimentalAttribute()
29. AWS_AUTH_REGION - testAuthRegionExperimentalAttribute()

- Tests leverage AWS SDK v2's getValueForField() API for clean, mockable attribute extraction
- Includes comprehensive testing for:
     - Core AWS services (S3, DynamoDB, SQS, SNS, Kinesis, Lambda, Step Functions, Secrets Manager)
     - Bedrock Gen AI attributes with JSON parsing validation
     - Bedrock resource attributes (Agent, Knowledge Base, Data Source)
     - Authentication attributes (access key, region)

### AWS SDK v1.11 (awssdk_v1_11)
New attributes being tested:
1. AWS_STREAM_ARN - testKinesisExperimentalAttributes() & contract tests
2. AWS_TABLE_ARN - testTableArnExperimentalAttributes() (Service identification only)
3. AWS_AGENT_ID - contract tests
4. AWS_KNOWLEDGE_BASE_ID - contract tests
5. AWS_DATA_SOURCE_ID - contract tests
6. AWS_GUARDRAIL_ID - testBedrockGuardrailAttributes() (Service identification only) & contract tests
7. AWS_GUARDRAIL_ARN - testBedrockGuardrailAttributes() (Service identification only) & contract tests
8. AWS_BEDROCK_RUNTIME_MODEL_ID -  testBedrockRuntimeAttributes() (Service identification only) & contract tests
9. AWS_BEDROCK_SYSTEM - contract tests
10. GEN_AI_REQUEST_MAX_TOKENS - contract tests
11. GEN_AI_REQUEST_TEMPERATURE - contract tests
12. GEN_AI_REQUEST_TOP_P - contract tests
13. GEN_AI_RESPONSE_FINISH_REASONS - contract tests
14. GEN_AI_USAGE_INPUT_TOKENS - contract tests
15. GEN_AI_USAGE_OUTPUT_TOKENS - contract tests
16. AWS_STATE_MACHINE_ARN - testStepFunctionsExperimentalAttributes() & contract tests
17. AWS_STEP_FUNCTIONS_ACTIVITY_ARN - contract tests
18. AWS_SNS_TOPIC_ARN - testSnsExperimentalAttributes()  & contract tests
19. AWS_SECRET_ARN - testSecretsManagerExperimentalAttributes() (Service identification only) & contract tests
20. AWS_LAMBDA_NAME - testLambdaNameExperimentalAttributes()
21. AWS_LAMBDA_ARN - testLambdaArnExperimentalAttributes()
22. AWS_LAMBDA_RESOURCE_ID - testLambdaResourceIdExperimentalAttributes() (Service identification only)
23. AWS_AUTH_ACCESS_KEY - testAuthAccessKeyAttributes()

*V1.11 is harder to test:*

V1.11 uses Java reflection to dynamically find and call methods like getFunctionName() on AWS request objects at runtime. This creates several testing challenges:
- Mock Method Mismatch: When you mock an AWS request object, it doesn't have the actual methods that reflection is trying to find. The reflection silently fails and returns null, making tests pass even though no attributes were extracted.
- Class Dependencies: To test properly, you'd need real AWS SDK classes instead of mocks, creating tight coupling between tests and external dependencies.
- Nested Object Complexity: Many attributes require traversing nested properties, which means mocking entire object graphs with proper method chains.

Contract tests sidestep these issues by using real AWS SDK objects against LocalStack, testing the complete end-to-end flow including actual reflection behavior without the complexity of mocking Java's reflection system.

### Related
- PRs for aws-sdk v1.11: #1115 and #1117 
- PRs for aws-sdk v2.2: #1111 and #1113
- Replaces patch: [current patch](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/patches/opentelemetry-java-instrumentation.patch)


.github/patches/versions
```diff
@@ -1,2 +1 @@
-OTEL_JAVA_INSTRUMENTATION_VERSION=v2.11.0
```
pxaws: Could you scanned the code base to find the places where the string `OTEL_JAVA_INSTRUMENTATION_VERSION` is used and make the changes accordingly?

.github/patches/opentelemetry-java-instrumentation.patch
pxaws, line 1: Since you delete the patch file, I think you might also need to change the places where it is used.

.github/patches/versions
anahatAWS: It is used in 3 places: 
1. .github/scripts/patch.sh ([here](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/scripts/patch.sh#L54) but not used if patch file (ie. .github/patches/opentelemetry-java-instrumentation.patch) doesn't exist)
2. scripts/local_patch.sh ([here](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/scripts/local_patch.sh#L70) but not used if patch file  doesn't exist)
3. lambda-layer/build-layer.sh ([here](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/lambda-layer/build-layer.sh#L10))

Because the lambda-layer uses that version for patching as well, I believe it's better to add it back in the original location. The other 2 patch scripts won't get/need this, but the lambda-layer needs it.

anahatAWS: It is used in 3 places: 
1. .github/scripts/patch.sh ([here](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/scripts/patch.sh#L54) but not used if patch file (ie. .github/patches/opentelemetry-java-instrumentation.patch) doesn't exist)
2. scripts/local_patch.sh ([here](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/scripts/local_patch.sh#L70) but not used if patch file  doesn't exist)
3. lambda-layer/build-layer.sh ([here](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/lambda-layer/build-layer.sh#L10))

Because the lambda-layer uses that version for patching as well, I believe it's better to add it back in the original location. The other 2 patch scripts won't get/need this, but the lambda-layer needs it.

.github/patches/opentelemetry-java-instrumentation.patch
anahatAWS: The only other places that use the patch file (other than .github/scripts/patch.sh and scripts/local_patch.sh) is the lambda-layer/build-layer.sh. I've removed the line where the instrumentation patching takes place [here](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/lambda-layer/build-layer.sh#L26) because the patch logic is applied when ADOT is built at the end of the script [here](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/lambda-layer/build-layer.sh#L54). 

Other than that, the other 2 scripts have safe guards to check if the instrumentation patch exists before trying to patch ([like this](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/scripts/local_patch.sh#L64)). But I will remove them all together as they are not needed in the future either.

anahatAWS:
Due to change in dependencyManagement otelVersion [here](https://github.com/aws-observability/aws-otel-java-instrumentation/pull/1120/files), lambda-layer patch file [aws-otel-java-instrumentation.patch](https://github.com/anahatAWS/aws-otel-java-instrumentation/blob/spi-final/lambda-layer/patches/aws-otel-java-instrumentation.patch#L9) has been updated. Also, because we removed the instrumentation patch that changed the upstream version.gradle.kts version names [here](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/patches/opentelemetry-java-instrumentation.patch#L4189), we need to update lambda-layer patch file [opentelemetry-java-instrumentation](https://github.com/anahatAWS/aws-otel-java-instrumentation/blob/spi-final/lambda-layer/patches/opentelemetry-java-instrumentation.patch#L315) (which builds off of original patch version) as well.

lambda-layer/patches/aws-otel-java-instrumentation.patch
```diff
@@ -3,11 +3,11 @@ index 9493189..6090207 100644
 --- a/dependencyManagement/build.gradle.kts
 +++ b/dependencyManagement/build.gradle.kts
 @@ -27,7 +27,7 @@ data class DependencySet(val group: String, val version: String, val modules: Li
- val TEST_SNAPSHOTS = rootProject.findProperty("testUpstreamSnapshots") == "true"
+ val testSnapshots = rootProject.findProperty("testUpstreamSnapshots") == "true"
 
  // This is the version of the upstream instrumentation BOM
--val otelVersion = "2.11.0-adot1"
+-val otelVersion = "2.11.0"
 +val otelVersion = "2.11.0-adot-lambda1"
```
pxaws, line 10: Any reason for the suffix `-adot-lambda1` in the string `2.11.0-adot-lambda1`?

lambda-layer/patches/opentelemetry-java-instrumentation.patch
pxaws, line 1: I am wondering why we need to patch the `version.gradle.kts`? Can we use the existing versions and potentially change the script to build lambda layer?

lambda-layer/patches/aws-otel-java-instrumentation.patch
pxaws, line 1: I am wondering why we need to patch `dependencyManagement/build.gradle.kts`? Can we just built it and then use whatever version in lambda build script?

lambda-layer/patches/opentelemetry-java-instrumentation.patch
anahatAWS: Addressed offline: this is patching for upstream

lambda-layer/patches/aws-otel-java-instrumentation.patch
anahatAWS: When we apply the lambda-layer [instrumentation patch](https://github.com/anahatAWS/aws-otel-java-instrumentation/blob/spi-final/lambda-layer/patches/aws-otel-java-instrumentation.patch#L10) in upstream, we label it as 2.11.0-adot-lambda1 and push to maven local when we are done. Later, when we built ADOT in the [build-layer.sh](https://github.com/anahatAWS/aws-otel-java-instrumentation/blob/spi-final/lambda-layer/build-layer.sh#L51), we need to search for 2.11.0-adot-lambda1 in maven  local, since we renamed it. That is why we patch `dependencyManagement/build.gradle.kts`. If we left it as 2.11.0, any other file expecting 2.11.0 in ADOT would get the lambda patches as well (not sure if that is the desired outcome). I think it makes sense to keep the current naming convention because it keeps the lambda code clean and separate from the rest of ADOT. Or else, I think there could be confusion as to which OTel version (patched or unpatched) we are using throughout the project.

lambda-layer/patches/aws-otel-java-instrumentation.patch
anahatAWS: Because we patch `version.gradle.kts` in upstream, rename to `2.11.0-adot-lambda1`, and push it to maven local, we need to update our `depencencyManagement` to search for the same project name.

.github/actions/patch-dependencies/action.yml
```diff
@@ -49,9 +49,6 @@ runs:
         if [[ -f .github/patches/opentelemetry-java.patch ]]; then
```
bjrara, line 49: Should we remove the patch step completely? We don't have any patches for the other repositories.

anahatAWS: It can be removed, but since the opentelemetry-java-instrumentation repo is the only one removed by the SPI, I left the other steps incase we add them in the future. I thought it would be an extra step to add back in for future patching on this repo. It also checks to see if the patch exists before trying to patch, so it shouldn't be an issue.

instrumentation/aws-sdk/build.gradle.kts
```diff
@@ -28,16 +28,17 @@ dependencies {
...
   compileOnly("software.amazon.awssdk:aws-core:2.2.0")
   compileOnly("software.amazon.awssdk:aws-json-protocol:2.2.0")
-
   compileOnly("net.bytebuddy:byte-buddy")
-  compileOnly("com.google.code.findbugs:jsr305:3.0.2")
 
   testImplementation("com.google.guava:guava")
-  testImplementation("io.opentelemetry.javaagent:opentelemetry-testing-common")
-
-  testImplementation("com.amazonaws:aws-java-sdk-core:1.11.0")
```
bjrara, line 38: It seems we completely removed the `com.amazonaws:aws-java-sdk-core:1.11.0`. Was it used before? Did we remove related tests?

.github/actions/patch-dependencies/action.yml
bjrara: Then why did we remove this step, we can simply remove the patch file, and won't cause any issue I guess?

anahatAWS: I removed it because we wouldn't need to add instrumentation patches in the future with this spi solution. But I can add it back in, just to be safe.

instrumentation/aws-sdk/build.gradle.kts
anahatAWS: It was for [AdotAwsSdkClientAdviceTest](https://github.com/anahatAWS/aws-otel-java-instrumentation/blob/spi-final/instrumentation/aws-sdk/src/test/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v1_11/AdotAwsSdkClientAdviceTest.java) before, but I updated to specific versions instead (to match the patch's version of aws sdk dependencies [here](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/patches/opentelemetry-java-instrumentation.patch#L40)) for [AwsSdkExperimentalAttributesInjectionTest](https://github.com/anahatAWS/aws-otel-java-instrumentation/blob/spi-final/instrumentation/aws-sdk/src/test/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v1_11/AwsSdkExperimentalAttributesInjectionTest.java).  No tests were effected by this change.

bjrara (approved):

bjrara (approved):

PR #1122, opened 2025-07-29
add version.gradle.kts

ezhang6811:
Issue #, if available:

Description of changes:
(copied from #1121)
The current version for SDK and Lambda Layer releases is only provided manually in the release workflow and auto-generated by the nebula release plugin based on previous tags, but never specified in the code. Adding a version file will provide explicit version management and make it easier to track upcoming releases. adotVersion is currently set to 2.11.1-dev0, following the convention used by our other ADOT projects ([Python](https://github.com/aws-observability/aws-otel-python-instrumentation/blob/main/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/version.py), [.NET](https://github.com/aws-observability/aws-otel-dotnet-instrumentation/blob/main/src/AWS.Distro.OpenTelemetry.AutoInstrumentation/Version.cs), [NodeJS](https://github.com/aws-observability/aws-otel-js-instrumentation/blob/629faf5d1608e6fd803f9b2c2e15d8c26ff7b40a/package.json#L3)).

The version provided in version.gradle.kts can still be overridden by release.version for uses cases such as building the lambda layer with the specified ADOT Java version.

Soon we will add pre-release and post-release workflows like the other language ADOT repos to automatically bump the version number. This is part of our work to align the release process for all 4 languages and streamline releases for future engineers.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -73,7 +73,7 @@ allprojects {
       ktlint("1.4.0").editorConfigOverride(mapOf("indent_size" to "2", "continuation_indent_size" to "2"))
 
       // Doesn't support pluginManagement block
-      targetExclude("settings.gradle.kts")
+      targetExclude("settings.gradle.kts", "version.gradle.kts")
 
       if (!project.path.startsWith(":sample-apps:")) {
         licenseHeaderFile("${rootProject.projectDir}/config/license/header.java", "plugins|include|import")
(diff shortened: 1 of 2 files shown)


srprash:
We have a version file here: https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/patches/versions

I believe it is currently used in building the lambda layer. Do we need to do something about that?

thpierce (approved):

PR #1127, opened 2025-08-05
Send main build metrics

ezhang6811:
*Issue #, if available:*

*Description of changes:*
Emit a failure metric if main build fails. Since this workflow is triggered with pushes to main or a release branch, we want to be notified if there is a failure with the build process or e2e tests.

Tested by temporarily adding an on: push: trigger to my own branch in this repo and testing the updated workflow. Verified that failure metric was successfully published to cloudwatch.
https://github.com/aws-observability/aws-otel-java-instrumentation/actions/runs/16783145872

### DIFF
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -265,7 +265,7 @@ jobs:
 
   publish-build-status:
     name: "Publish Main Build Status"
-    needs: [ build, contract-tests, application-signals-e2e-test ]
+    needs: [ build, e2e-test, contract-tests, application-signals-lambda-layer-build, application-signals-e2e-test ]
     runs-on: ubuntu-latest
     if: always()
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/main-build.yml
```diff
@@ -263,16 +263,22 @@ jobs:
       adot-image-name: ${{ needs.build.outputs.staging-image }}
 
   publish-build-status:
-    needs: [ build, contract-tests ]
-    if: ${{ always() }}
-    uses: ./.github/workflows/publish-status.yml
```
srprash, line 268: The publish-status workflow already publishes some metrics around the main build. I did not look into it much, but want to understand why does it currently not work for us and why do we need this change?

ezhang6811: The workflow publishes both a success and failure metric. For the purposes of our alarms, we only need a failure metric for main build. The workflow is still used by other e2e tests

wangzlei (approved):

PR #1141, opened 2025-08-12
AWS X-Ray Adaptive Sampling Support

majanjua-amzn:
### Description
AWS X-Ray will soon be supporting adaptive sampling through the sampling rule APIs, allowing customers to configure anomaly rate based sampling rate boosts. GetSamplingTargets and GetSamplingRules will be updated to support new inputs and outputs relevant to this feature, and as such the SDK must be updated.

#### Sampling boost
The SDK now supports sending "boost statistics" to the GetSamplingTargets API. These statistics include the number of total requests (traces), the number of traces with anomalies detected (more on this later), and the number of anomalies sampled. The server then responds with instructions on what sampling rate to set and for how long. The SDK adjusts accordingly

#### Local configuration
Customer's also need to be able to define what an anomaly is for their applications to effectively provide boost statistics. By default, any 5XX error (or fallback to ERROR attribute) is treated as an anomaly. If provided, a local configuration can define specific criteria including error code regex, operation, and high latency threshold to count statistics based on.

#### Anomaly Capture (disabled by default)
Anomalies can also be captured directly when left unsampled. When anomalous spans are detected, a reservoir-style span capturing mechanism configured through the above local configuration will send the span directly to the spanExporter. These will appear in the console as partial traces and ensure the customer can see spans even if the boosted sampling rate was unable to capture the anomalies.

#### AWS X-Ray component patch update for OTel Java Contrib - see the following diff between the changes here and the release of the contrib we currently consume: [link](https://github.com/open-telemetry/opentelemetry-java-contrib/compare/release/v1.39.x...majanjua-amzn:opentelemetry-java-contrib:adaptive-sampling) (includes diff from previous patch on the sampler)
- Added an import for OTel semantic conventions
- Created a class called `AwsSamplingResult` that includes the matched sampling rule in the trace state or propagates the received sampling rule from an upstream call using the trace state
- Added a class called `AwsXrayAdaptiveSamplingConfig` for the local SDK configuration option
- Added the config object and a batch span exporter to the `AwsXrayRemoteSampler` to allow identification and export of anomalies
- Added `adaptSampling` function that is called on each span and acts if and only if adaptive sampling configurations are present - this is where the core logic of the feature is
- Updated calls to GetSamplingTargets to include boost statistics
- Updated GetSamplingRules and GetSamplingTargets request and response classes where relevant
- In `SamplingRuleApplier`:
  - Able to receive boost information
  - Able to receive boost related statistics from the `XrayRulesSampler`
  - Fixed bug where sampling rules are scheduled to call GetSamplingTargets at slightly different times
- In `XrayRulesSampler`:
  - Added attribute to spans when boost config/anomaly capturing is enabled to be able to identify boost-enabled systems in X-Ray backend
  - Accept `AwsXrayAdaptiveSamplingConfig` and apply it in `adaptSampling` to change anomaly capturing/boost logic
  - Get and propagate upstream sampling rule in `shouldSample` using `AwsSamplingResult`
  - Core implementation of `adaptSampling`
    - If account has no boost, return
    - [1] Gets rule to report to based on upstream matched rule propagated through trace state, [2] identifies anomalies based on local config or default 5XX, [3] captures anomaly if error capture is enabled, [4] counts boost statistics
    - Maintain `anomalyTracesSet` that holds trace IDs for anomaly spans to ensure we don't double count anomalies in one trace. When the local root span for this trace is encountered, it is removed from the set
  - Add `generateIngressOperation` based on ADOT SDK logic for getting operation - used for matching with operations provided in local configuration
- Add unit tests

#### ADOT SDK Changes
- Add YAML import for reading local configuration
- Remove B3 and multi propagators as they remove/override the sampling rule propagated through trace state and are no longer needed
- Update `customizeSampler` to provide the sampler the span exporter and the local adaptive sampling configuration and pass the sampler to the span metrics processor
- Call `adaptSampling` on each span from the span metrics processor
- Add parsing function and associated test

### Testing
- Automated release test passing and going through PR here: [aws-application-signals-test-framework#442](https://github.com/aws-observability/aws-application-signals-test-framework/pull/442)
- Manual testing done using 3 services, A -> B -> C, where A has a boosted sampling rule and B and C produce anomalies that are sent to the backend, boosting the sampling rate in A


.github/patches/opentelemetry-java-contrib.patch
```diff
@@ -1,11 +1,564 @@
+diff --git a/aws-xray/build.gradle.kts b/aws-xray/build.gradle.kts
+index ccec9d52..f764bba9 100644
+--- a/aws-xray/build.gradle.kts
++++ b/aws-xray/build.gradle.kts
```
lukeina2z, line 4: The X-Ray remote sampler is a small library. Instead of applying the patch here, I believe the right approach is to add a new library in this repo by copying the code from https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/aws-xray. Build and release it from this repo, and then contribute your code changes upstream.

Did you submit this change to upstream?

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsSpanMetricsProcessorTest.java
```diff
@@ -112,7 +116,7 @@ public void testStartDoesNothingToSpan() {
     Context parentContextMock = mock(Context.class);
     ReadWriteSpan spanMock = mock(ReadWriteSpan.class);
     awsSpanMetricsProcessor.onStart(parentContextMock, spanMock);
-    verifyNoInteractions(parentContextMock, spanMock);
+    //    verifyNoInteractions(parentContextMock, spanMock);
```
lukeina2z: typo?

.github/patches/opentelemetry-java-contrib.patch
majanjua-amzn: I disagree with the first part, so far it's been standard practice for us to patch if it is needed. The same was done for the previous sampler patch change #1089 

Code duplication is not a reasonable alternative to this.

I will be submitting this change to upstream once this PR has gone through 1-2 rounds of review so I don't have to update both PRs repeatedly.

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsSpanMetricsProcessorTest.java
majanjua-amzn: Good catch, fixed

wangzlei (approved):

wangzlei (approved):

PR #1156, opened 2025-08-19
fix(deps): update OTel Java Agent to v2.18.1 and OTel Contrib to v1.48.0

lukeina2z:
This updates upstream dependency OTel Java Agent to v2.18.1 and upgrades upstream OTel Contrib to v1.48.0.

It removes two patch files, as those code has been migrated to upstream.

For the OTel Java Contrib patch, it has been replaced by PR-1959 https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1959
, which is included in the OTel Java Contrib v1.48.0 release.

For the OTel Java Agent patch for the Lambda stream handler, it has been replaced by PR-13466 https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/13466
, which is included in the OTel Java Agent v2.18.1 release.

Tests:
- Unit tests pass: ./gradlew build test
- Smoke/Contract tests pass: ./gradlew appsignals-tests:contract-tests:contractTests
- X-Ray remote sampling service end-to-end manual tests pass
- EC2 end-to-end tests with SpringBoot Java app pass
- Lambda end-to-end tests with SpringBoot Java app pass


.github/patches/versions
thpierce, line 1: Are we able to actually remove this file? We have no patches today, right? Does this file get used for anything else?

thpierce, line 1: Are we able to actually remove this file and/or remove `OTEL_JAVA_INSTRUMENTATION_VERSION=v2.18.1`? We have no patches today, right? Does this file get used for anything else?

lambda-layer/build-layer.sh
```diff
@@ -25,28 +25,10 @@ git checkout v${version} -b tag-v${version}
 # This patch is for Lambda related context propagation
 patch -p1 < "$SOURCEDIR"/patches/opentelemetry-java-instrumentation.patch
 
-patch -p1 < "$SOURCEDIR"/patches/StreamHandlerInstrumentation.patch
```
thpierce, line 28: Can you explain why it's safe to remove this? Has our patched logic been contributed to upstream? If so, can you just link to the upstream PR that merged this, so I can verify we are ok to remove this logic?

lambda-layer/build-layer.sh
```diff
@@ -25,28 +25,10 @@ git checkout v${version} -b tag-v${version}
...
-  pushd opentelemetry-java-contrib
-  git checkout v${contrib_version} -b tag-v${contrib_version}
-
-  # There is another patch in the .github/patches directory for other changes. We should apply them too for consistency.
-  patch -p1 < "$SOURCEDIR"/../.github/patches/opentelemetry-java-contrib.patch
-
-  ./gradlew publishToMavenLocal
-  popd
-  rm -rf opentelemetry-java-contrib
-fi
```
thpierce: If I'm reading this right, we don't do anything as long as `version` doesn't contain `OTEL_JAVA_CONTRIB_VERSION`. I wonder if we should leave this in place, in case we need to add patches in future? Just to avoid re-inventing the wheel.

build.gradle.kts
```diff
@@ -108,6 +108,7 @@ allprojects {
       testImplementation("org.junit.jupiter:junit-jupiter-api")
       testImplementation("org.junit.jupiter:junit-jupiter-params")
       testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine")
+      testRuntimeOnly("org.junit.platform:junit-platform-launcher")
```
thpierce: Curious - why is this needed now?

.github/patches/versions
```diff
@@ -1,2 +1 @@
-OTEL_JAVA_INSTRUMENTATION_VERSION=v2.11.0
-OTEL_JAVA_CONTRIB_VERSION=v1.39.0
```
thpierce, line 2: Can you explain why it's safe to remove this? Has our patched logic been contributed to upstream? If so, can you just link to the upstream PR that merged this, so I can verify we are ok to remove this logic?

.github/patches/versions
lukeina2z: We cannot remove it right now. Lambda build still need local patch.

lambda-layer/build-layer.sh
lukeina2z: Upstream has the fix:
feat: Add instrumentation for Lambda Java interface HandleStreamRequest #13466
https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/13466

lambda-layer/build-layer.sh
lukeina2z: Will add it back.

build.gradle.kts
lukeina2z: This is required by the SPI project Vancouver team added. I will move this line to that project's specific build file. 

I believe SPI introduced a new dependency from upstream. Upstream made a change this Feb. which is causing this: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/13372

.github/patches/versions
lukeina2z: This PR does not remove OTel Java Agent patch, Vancouver team removed it. Please talk to SPI project owner for details.

For OTel Contrib patch, the only thing that patch did is the Remote Sampler fix Mo made. Upstream v1.48 will cover it. This PR will adopt OTel Contrib v1.48. So it's safe to delete the patch. The upstream fix: https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1959

lambda-layer/patches/StreamHandlerInstrumentation.patch
```diff
@@ -1,513 +0,0 @@
...
-+        messageScope.close();
-+        AwsLambdaInstrumentationHelper.messageInstrumenter()
-+            .end(messageContext, (SQSEvent) arg, null, throwable);
-+      }
-+
-+      if (functionScope != null) {
-+        functionScope.close();
-+        AwsLambdaInstrumentationHelper.functionInstrumenter()
-+            .end(functionContext, input, null, throwable);
-+      }
```
thpierce, line 389: This appears to be different from what was contributed to the [upstream](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-lambda/aws-lambda-events-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awslambdaevents/v2_2/AwsLambdaRequestStreamHandlerInstrumentation.java#L55-L84). Can you explain the discrepancy?

lukeina2z: I only saw three variables have different names:

One introduced by upstream two weeks ago:

Rename lambda instrumentation helper class #14371
https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/14371/files

Another two are, in the patch we call it "functionContext" defined in line 350, "functionScope" defined in line 351.  In upstream, we call it "otelContext" and "otelScope".

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.18.x/instrumentation/aws-lambda/aws-lambda-events-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awslambdaevents/v2_2/AwsLambdaRequestStreamHandlerInstrumentation.java#L60-L61

After investigation, I believe the upstream behavior matches the patch.

thpierce: In this patch `stopSpan` looks at `messageScope` and `functionScope`, but in the [upstream](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.18.x/instrumentation/aws-lambda/aws-lambda-events-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awslambdaevents/v2_2/AwsLambdaRequestStreamHandlerInstrumentation.java#L83-L87), we only handle `functionScope`. Am I missing something?

lukeina2z: I think I now understand why the upstream implementation differs from the patch.

The upstream version is correct. The patch contains redundant code, not bug, but we should use the upstream code.

In the patch, "messageScope" is always null, so that "if" block never executes:
https://github.com/aws-observability/aws-otel-java-instrumentation/blob/release/v2.11.x/lambda-layer/patches/StreamHandlerInstrumentation.patch#L379-L383

When a Lambda function is triggered by an SQS event, it does not use the StreamHandler interface. With StreamHandler, we cannot assume the argument is an SQSEvent, it’s just a stream.

For SQSEvent, Lambda uses a non-Stream handler, as shown here:
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.18.x/instrumentation/aws-lambda/aws-lambda-events-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awslambdaevents/v2_2/AwsLambdaRequestHandlerInstrumentation.java#L84-L90

So SQS Event should be removed from the Stream handler.

lukeina2z: line 382 In the patch (also the quote above) will never run, otherwise that will crash.

+            .end(messageContext, (SQSEvent) arg, null, throwable);

'arg' is 'InputStream' type, we can't cast it to SQSEvent. That's wrong.

.github/patches/opentelemetry-java-contrib.patch
```diff
@@ -1,237 +0,0 @@
...
-   mustRunAfter("jar")
- }
- 
-+tasks.withType<Test>().configureEach {
-+  dependsOn("shadowJar")
-+}
-+
- // The javadoc from wire's generated classes has errors that make the task that generates the "javadoc" artifact to fail. This
- // makes the javadoc task to ignore those generated classes.
- tasks.withType(Javadoc::class.java) {
```
srprash, line 225: What about this patch for `disk-buffering`. What is it used for and can it be safely removed?

lambda-layer/patches/aws-otel-java-instrumentation.patch
```diff
@@ -1,13 +1,13 @@
 diff --git a/dependencyManagement/build.gradle.kts b/dependencyManagement/build.gradle.kts
-index 9493189..6090207 100644
+index d186406..91b9386 100644
 --- a/dependencyManagement/build.gradle.kts
 +++ b/dependencyManagement/build.gradle.kts
 @@ -27,7 +27,7 @@ data class DependencySet(val group: String, val version: String, val modules: Li
  val testSnapshots = rootProject.findProperty("testUpstreamSnapshots") == "true"
-
+
```
srprash, line 7: nit: intentional white space?

lambda-layer/build-layer.sh
srprash: Synced offline with Luke to understand this change. Re: removing the `SQSEvent` handling, he explained 2 things:
1. Due to an existing bug, the functionality never worked.
2. The handling doesn't need to be fixed and can be safely removed because the `InputStream` argument in the [`RequestStreamHandler`](https://www.javadoc.io/doc/com.amazonaws/aws-lambda-java-core/1.2.0/com/amazonaws/services/lambda/runtime/RequestStreamHandler.html) will never be of type `SQSEvent`. Which means that this instrumentation does not need to handle the `SQSEvent` specifically.

lambda-layer/patches/aws-otel-java-instrumentation.patch
lukeina2z: It's a copy/paste from git diff > mydiff.txt. I want to keep everything generated by that command so the checksum on line 2 won't be broken.

lukeina2z: It seems the original file always has the blank line there. I don't know why the previous patch did not have it.   https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/dependencyManagement/build.gradle.kts#L24-L30

.github/patches/opentelemetry-java-contrib.patch
lukeina2z: It's a folder in OTel Java Contrib repo:  https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/disk-buffering

Mo has detailed how he created this patch:

Added patch for the X-Ray Remote Sampler #1089
https://github.com/aws-observability/aws-otel-java-instrumentation/pull/1089

The only thing added by the patch is to build fat jar for disk-buffering library. I don't think it has impact on ADOT Java Agent. It might be an error made by the patch maker. A local change for testing purpose accidentally included in the patch?

lambda-layer/patches/aws-otel-java-instrumentation.patch
srprash: Ack

.github/patches/opentelemetry-java-contrib.patch
srprash: > It might be an error made by the patch maker. A local change for testing purpose accidentally included in the patch?

Can you confirm with @AsakerMohd ?

srprash (approved):

PR #1159, opened 2025-08-20
fix(deps): replace deprecated Gradle wrapper validation action

lukeina2z:
The latest release of gradle/wrapper-validation-action is v3.5, but ADOT Java is currently using v1.

This PR replaces the deprecated gradle/wrapper-validation-action with the new gradle/actions/wrapper-validation action. v4.4.2 is an official release. It's commit hash is #017a9ef.
https://github.com/gradle/actions/releases/tag/v4.4.2

When uses third-party action, we should always use commit hash instead of release number, for security and integrity reason. https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

References:
- https://github.com/gradle/wrapper-validation-action
- https://github.com/gradle/actions/releases/


.github/workflows/codeql-analysis.yml
```diff
@@ -50,7 +50,7 @@ jobs:
         gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
         gpg_password: ${{ secrets.GPG_PASSPHRASE }}
         
-    - uses: gradle/wrapper-validation-action@v1
+    - uses: gradle/actions/wrapper-validation@017a9effdb900e5b5b2fddfb590a105619dca3c3 # v4.4.2
```
thpierce, line 53: Why do we not just use `v3`? I don't really love depending on a specific (unreleased?) commit unless we have a good reason for it.

thpierce (approved):

.github/workflows/codeql-analysis.yml
lukeina2z: v3 is too old. v4.4.2 is an official release. It's commit hash is #017a9ef.  
https://github.com/gradle/actions/releases/tag/v4.4.2

When uses third-party action, we should always use commit hash instead of release number, for security and integrity reason. https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

thpierce: Question: Will dependabot be able to update gradle/actions/wrapper-validation if we pin to commitID? I'd like to rely on it for keeping things up to date going forward (I know it didn't keep v1 up to date, but I think someone must have manually done that).

Answer: For each action in the file, Dependabot checks the action's reference (typically a version number or commit identifier associated with the action
https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

thpierce (approved):

PR #1162, opened 2025-08-22
[semconv]: Add support for new formal DB semantic convention keys

lukeina2z:
ADOT Java currently uses some deprecated incubating semconv keys. This PR adds support for the newly introduced formal semconv keys that replace them, while maintaining backward compatibility by falling back to the legacy keys when necessary.

**Deprecated keys:**
- DB_NAME
- DB_OPERATION
- DB_STATEMENT
- DB_SYSTEM

(Reference: https://github.com/open-telemetry/semantic-conventions-java/blob/release/v1.34.0/semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/DbIncubatingAttributes.java#L322-L327)

**New keys:**
- DB_NAMESPACE
- DB_OPERATION_NAME
- DB_QUERY_TEXT
- DB_SYSTEM_NAME

**Tests performed:**
- Unit tests: `./gradlew build test`
- Smoke/contract tests: `./gradlew appsignals-tests:contract-tests:contractTests`
- Manual E2E test with SpringBoot sample app.

### DIFF
--- a/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
+++ b/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
@@ -15,6 +15,10 @@
 
 package software.amazon.opentelemetry.javaagent.providers;
 
+import static io.opentelemetry.semconv.DbAttributes.DB_NAMESPACE;
+import static io.opentelemetry.semconv.DbAttributes.DB_OPERATION_NAME;
+import static io.opentelemetry.semconv.DbAttributes.DB_QUERY_TEXT;
+import static io.opentelemetry.semconv.DbAttributes.DB_SYSTEM_NAME;
 import static io.opentelemetry.semconv.HttpAttributes.HTTP_REQUEST_METHOD;
 import static io.opentelemetry.semconv.HttpAttributes.HTTP_RESPONSE_STATUS_CODE;
 import static io.opentelemetry.semconv.NetworkAttributes.NETWORK_PEER_ADDRESS;
@@ -26,7 +30,6 @@ import static io.opentelemetry.semconv.UrlAttributes.URL_FULL;
 // https://github.com/open-telemetry/semantic-conventions-java/blob/release/v1.34.0/semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/DbIncubatingAttributes.java#L322-L327
 // They have been replaced with new keys:
 // https://github.com/open-telemetry/semantic-conventions-java/blob/release/v1.34.0/semconv/src/main/java/io/opentelemetry/semconv/DbAttributes.java#L77
-// TODO: Supporting new keys. Cannot do this now as new keys are not available in OTel Agent 2.11.
 // TODO: Delete deprecated keys once they no longer exist in binding version of the upstream code.
 import static io.opentelemetry.semconv.incubating.DbIncubatingAttributes.DB_CONNECTION_STRING;
 import static io.opentelemetry.semconv.incubating.DbIncubatingAttributes.DB_NAME;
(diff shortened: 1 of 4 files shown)


awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -349,10 +354,7 @@ private static void setRemoteEnvironment(SpanData span, AttributesBuilder builde
   private static String generateRemoteOperation(SpanData span) {
     String remoteOperation = UNKNOWN_REMOTE_OPERATION;
     if (isKeyPresent(span, URL_FULL) || isKeyPresent(span, HTTP_URL)) {
```
thpierce: Not blocking: If you're changing the below code, may as well update this code to use `isKeyPresentWithFallback`.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -949,6 +948,15 @@ private static String getRemoteService(SpanData span, AttributeKey<String> remot
     return remoteService;
   }
 
+  static String getRemoteServiceWithFallback(
```
thpierce, line 951: Not blocking: you could just update the signature of `getRemoteService` to be `getRemoteService(SpanData span, AttributeKey<String>... remoteServiceKeys)` and have it iterate through all keys until there is a hit. I imagine a future where we have to support 3+ different keys, given how long it takes upstream to update.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
```diff
@@ -949,6 +948,15 @@ private static String getRemoteService(SpanData span, AttributeKey<String> remot
...
+      SpanData span, AttributeKey<String> remoteSvcKey, AttributeKey<String> remoteSvcFallbackKey) {
```
thpierce: Lets try to follow class convention, don't abbreviate:
```suggestion
      SpanData span, AttributeKey<String> remoteServiceKey, AttributeKey<String> remoteServiceFallbackKey) {
```

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
thpierce, line 1: The tests in this class are intentionally robust as the logic is complex and we really want to be sure what it is doing. Lets take the effort and update relevant tests with these new variables. Namely:

* testGetDBStatementRemoteOperation - add case that shows DB_QUERY_TEXT takes priority over DB_STATEMENT but not DB_OPERATION
* testDBClientSpanWithRemoteResourceAttributes - add case for DB_NAMESPACE
* testRemoteAttributesCombinations - add cases for DB_SYSTEM_NAME/DB_OPERATION_NAME

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
lukeina2z: Updated code.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
lukeina2z: Done.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
lukeina2z: I prefer not to use a varargs parameter for this method, for three main reasons:
1. Varargs allows zero or more arguments, but in this case, we clearly expect at least two. Allowing zero arguments introduces unnecessary ambiguity and potential misuse.
2. Single Responsibility Principle. The core function should handle a single, well-defined task: retrieving a remote service for one specific key. Fallback logic should be implemented in a separate method.
3. Readability and clarity: The function signature should make its purpose and behavior immediately clear. 

If we need to handle multiple fallback keys in the future, we can introduce a new overload that accepts a list of fallback keys:
```java
getRemoteServiceWithFallback(
    SpanData span,
    AttributeKey<String> remoteServiceKey,
    List<AttributeKey<String>> remoteServiceFallbackKeys
)
```

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGeneratorTest.java
lukeina2z: New tests added.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java
thpierce: Ok, not blocking.

thpierce (approved):

PR #1163, opened 2025-08-25
Merge SDK and Lambda Releases

ezhang6811:
*Issue #, if available:*

*Description of changes:*
We plan to consolidate our ADOT SDK and Lambda layer releases for future versions. This PR merges the Lambda release workflow into the main release workflow, and publishes the layer artifacts and ARN notes to the same Github release as the SDK.

The release build workflow also now includes the SDK release notes with our upstream dependency versions to eliminate the manual effort needed when updating release notes.


.github/workflows/release-build.yml
```diff
@@ -55,23 +62,67 @@ jobs:
           gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
           gpg_password: ${{ secrets.GPG_PASSPHRASE }}
 
-      - name: Configure AWS Credentials
-        uses: aws-actions/configure-aws-credentials@v4
+      - name: Build release with Gradle
+        uses: gradle/gradle-build-action@v3
         with:
-          role-to-assume: ${{ secrets.AWS_ASSUME_ROLE_ARN }}
-          aws-region: ${{ env.AWS_PUBLIC_ECR_REGION }}
```
thpierce, line 62: Why is it ok to no longer sign in with `${{ secrets.AWS_ASSUME_ROLE_ARN }}`?

.github/workflows/release-build.yml
```diff
@@ -55,23 +62,67 @@ jobs:
...
+        run: |
+          ./build-layer.sh
+
+      - name: Upload layer
+        uses: actions/upload-artifact@v4
+        with:
+          name: layer.zip
+          path: lambda-layer/build/distributions/aws-opentelemetry-java-layer.zip
+
+  publish-sdk:
```
thpierce, line 125: Concern: Do we need to run `Publish patched dependencies to maven local` in this job before running `Build image for testing`. Suspect so.

.github/workflows/release-build.yml
```diff
@@ -137,28 +188,223 @@ jobs:
...
+      
+      - name: Download layer.zip artifact
+        uses: actions/download-artifact@v5
+        with:
+          name: layer.zip
+
+      - name: Rename artifacts
+        run: |
+          cp "aws-opentelemetry-agent-${{ github.event.inputs.version }}.jar" ${{ env.ARTIFACT_NAME }}
+          cp aws-opentelemetry-java-layer.zip layer.zip
```
thpierce, line 403: Is this needed if we just downloaded the artifacts with these names?

.github/workflows/release-lambda.yml
thpierce, line 1: Should we not delete this file?

.github/workflows/release-build.yml
```diff
@@ -137,28 +188,223 @@ jobs:
...
+
+      - name: download layer.zip
+        uses: actions/download-artifact@v5
+        with:
+          name: layer.zip
+
+      - name: publish
+        run: |
+          aws s3 mb s3://${{ env.BUCKET_NAME }}
+          aws s3 cp aws-opentelemetry-java-layer.zip s3://${{ env.BUCKET_NAME }}
```
thpierce, line 273: This is wrongly named now? Same with `--content S3Bucket=${{ env.BUCKET_NAME }},S3Key=aws-opentelemetry-java-layer.zip \`

.github/workflows/release-build.yml
```diff
@@ -137,28 +188,223 @@ jobs:
...
+          # Extract versions from dependency files
+          OTEL_INSTRUMENTATION_VERSION=$(grep "val otelVersion" dependencyManagement/build.gradle.kts | sed 's/.*= "\([^"]*\)".*/\1/')
+          OTEL_CONTRIB_VERSION=$(grep "io.opentelemetry.contrib:opentelemetry-aws-xray" dependencyManagement/build.gradle.kts | sed 's/.*:\([^"]*\)".*/\1/')
+
+          # Create release notes
+          cat > release_notes.md << EOF
+          This release contains updates of the following upstream components:
+
+          OpenTelemetry Java Contrib - $OTEL_CONTRIB_VERSION
+          Opentelemetry Instrumentation for Java - $OTEL_INSTRUMENTATION_VERSION
```
thpierce, line 420: Not blocking: We should probably do more like what JS does, and list more out. Out of scope for now.

.github/workflows/release-build.yml
ezhang6811: the artifact is uploaded/downloaded with name `layer.zip`, but it still retains the original file name. see [Python](https://github.com/aws-observability/aws-otel-python-instrumentation/blob/7e79092d034fbbf012792395e28de866ea877fd4/.github/workflows/release-lambda.yml#L94).

.github/workflows/release-build.yml
ezhang6811: see [response to this comment](https://github.com/aws-observability/aws-otel-java-instrumentation/pull/1163/files#r2342142029)

.github/workflows/release-lambda.yml
ezhang6811: yes, deleting.

.github/workflows/release-build.yml
ezhang6811: It looks like it is needed for the integration tests. adding back. to this job.

.github/workflows/release-build.yml
ezhang6811: as discussed, moved the steps `Publish patched dependencies to maven local`, `Configure AWS Credentials`, `Log in to AWS ECR`, and `Build release with Gradle` into the **publish-sdk** job as well so that the Dockerfile for building the image can reference the built artifact in the correct path.

.github/workflows/release-build.yml
thpierce: Oh man that is so so stupid... Ty.

thpierce (dismissed):
Do not merge until new release process is ready for release.

PR #1171, opened 2025-08-29
fix: UDP Exporter Sample App not starting via Gradle

jj22ee:
*Issue #, if available:*
- Sample app for UDP Exporter is failing via `gradle build`: https://github.com/aws-observability/aws-otel-java-instrumentation/actions/runs/17167485805/job/48710900750

*Description of changes:*
- UDP e2e test uses `gradle` from GH Action env, which isn't stable. Use `gradlew` that exists in this repo instead.


.github/workflows/udp-exporter-e2e-test.yml
```diff
@@ -51,8 +51,8 @@ jobs:
         run: |
           export XRAY_UDP_SPAN_EXPORTER_VERSION=${{ steps.build-udp-exporter.outputs.xrayUdpSpanExporterVersion }}
           echo "Running Sample App against X-Ray UDP Span Exporter version: $XRAY_UDP_SPAN_EXPORTER_VERSION"
-          gradle build
-          gradle bootRun &
+          ../../gradlew build
+          ../../gradlew bootRun &
```
lukeina2z, line 55: What's the reason not explicitly declare the project name and action when use gradlew?

./gradlew :sample-apps:spark-awssdkv1:build

jj22ee: I believe Colon syntax cannot be used because this UDP Exporter Sample App is not defined as a "subproject" in the root `settings.gradle.kts`. This is because is its own standalone Gradle project Sample App, and not intended to be tied directly to the root gradle build of ADOT.

PR #1191, opened 2025-09-16
Support Trace Context extraction from Lambda Context object, and respect user-configured OTEL_PROPAGATORS

jj22ee:
*Issue #, if available:*
- In latest `com.amazonaws:aws-lambda-java-core:1.4.0`, Lambda Context has a new `lambdaContext.getXrayTraceId()` method. We need to use this over SystemProperty/EnvVar to support multi-concurrency in Lambda.

*Description of changes:*
- respect OTEL_PROPAGATORS Env Var priority
  - Priority logic BEFORE:
    - Create `carrierA` to contain headers from Lambda request http headers and the custom client context
    - get X-Ray Trace ID from (in order or priority) SystemProperty or EnvVar, add it to new `carrierB`
    - Use Global Propagator to extract trace context from `carrierB` (trace extraction only works if xray propagator is configured)
    - If above trace extraction results in a context that is not valid&sampled, try again with Global Propagator using `carrierA`
  - Priority logic AFTER:
    - Create `carrierA` to contain headers from Lambda request http headers and the custom client context
    - get X-Ray Trace ID from (in order or priority) Lambda Context, SystemProperty, or EnvVar, add it to `carrierA` (will overwrite x-ray header value if present from Lambda request http headers).
    - Use Global Propagator to extract trace context from `carrierA`


wangzlei (dismissed):
LGTM

wangzlei (approved):

PR #1206, opened 2025-09-19
fix: Remove Netty as an agent dependency

lukeina2z:
Netty is currently bundled in the ADOT Java Agent JAR, unnecessarily increasing its size by ~7 MB.

This PR removes Netty from the ADOT Java Agent dependency list by:

1. Explicitly removing the Netty BOM.
2. Upgrading the AWS SDK to 2.33.11, which addresses the Netty security risk.
3. Replacing the inclusion of all AWS SDK packages with only the specific modules required by the ADOT Java Agent.

Tests performed:

Local build: ./gradlew build ✅
Unit tests: ./gradlew test ✅
Smoke/contract tests: ./gradlew appsignals-tests:contract-tests:contractTests ✅


awsagentprovider/build.gradle.kts
```diff
@@ -50,7 +50,6 @@ dependencies {
   compileOnly("io.opentelemetry:opentelemetry-exporter-otlp-common")
 
   // For OtlpAwsExporter SigV4 Authentication
-  runtimeOnly("software.amazon.awssdk:sts")
```
liustve, line 53: This dependency cannot be removed. It is required for customers using the Sigv4 exporters: https://github.com/aws-observability/aws-otel-java-instrumentation/pull/1101

dependencyManagement/build.gradle.kts
```diff
@@ -103,6 +98,9 @@ dependencies {
     for (dependency in dependencyLists) {
       api(dependency)
     }
+
+    api("software.amazon.awssdk:auth:2.33.11")
+    api("software.amazon.awssdk:aws-core:2.33.11")
```
liustve, line 103: Does adding `aws-core` dependency include `sts`? If it does can ignore my previous comment. Also would you help test the Sigv4 path to ensure that `sts` still works?

awsagentprovider/build.gradle.kts
lukeina2z: Why the build and Unit Test not fail after removing it? Do you know which API Agent uses?

dependencyManagement/build.gradle.kts
lukeina2z: Do we have any UTs for Sigv4 work? Core does not  include STS.

awsagentprovider/build.gradle.kts
liustve: This dependency is managed by the AWS SDK and is used internally in the the AWS SDK authentication libraries to retrieve credentials from all sources. This is already well tested by upstream which is why we don't have unit tests here. 

To validate this you have to just check one of these paths work:
https://docs.aws.amazon.com/sdkref/latest/guide/feature-assume-role-credentials.html#feature-assume-role-credentials-settings

dependencyManagement/build.gradle.kts
liustve: see above

awsagentprovider/build.gradle.kts
lukeina2z: I tested this PR with this document. Both trace and log are working.

Exporting collector-less telemetry using AWS Distro for OpenTelemetry (ADOT) SDK
https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-OTLP-UsingADOT.html

May I ask why you believe the SigV4 projects need STS? Are you trying to obtain an AWS access token from STS? If so, which AWS account, user, or role are you using? I wasn’t able to find the code.

liustve: Thank you, Luke! Yes, SigV4 requires STS. This request comes from users who rely on this feature but are unable to explicitly set or hardcode their credentials in their environment. Instead, these environments use the STS library to automatically retrieve and authenticate credentials on their behalf.

liustve: see: https://github.com/aws-observability/aws-otel-java-instrumentation/pull/1101

PR #1208, opened 2025-09-19
Fix: Incorrect zip file name in Lambda Java release workflow

lukeina2z:
Problem:
The Lambda Java layer release fails because the workflow uses the wrong artifact (layer zip file) name. https://github.com/aws-observability/aws-otel-java-instrumentation/actions/runs/17867947751

Solution:
Replaced layer.zip with the correct build artifact name: aws-opentelemetry-java-layer.zip.


.github/workflows/release-lambda.yml
```diff
@@ -213,7 +213,7 @@ jobs:
       - name: download layer.zip
         uses: actions/download-artifact@v4
         with:
-          name: layer.zip
+          name: aws-opentelemetry-java-layer.zip
```
thpierce, line 216: I think there may be some significance to the name `layer.zip` (may be mentioned in public docs, etc), can we just add a step to rename the file to `layer.zip` after downloading it?

thpierce (approved):

PR #1218, opened 2025-09-23
Fix Trace Context extraction from Lambda Context object by bypassing Muzzle Check

jj22ee:
*Issue #, if available:*
- https://github.com/aws-observability/aws-otel-java-instrumentation/actions/runs/17954879038/job/51066448865

AWS Lambda Environment may or may not have the `getXrayTraceId` method in Lambda Context (from Lambda Core dependency). This will cause OTel's Muzzle Check to disable the AWS Lambda Instrumentation in the event that it isn't present, causing no Lambda Function Handler Span to appear

Fix of previous PR: https://github.com/aws-observability/aws-otel-java-instrumentation/pull/1191

*Description of changes:*
- Use `@NoMuzzle` annotation when accessing `getXrayTraceId` in Lambda Context Object, which avoids OTel's muzzle check. If `NoSuchMethodError` is caught, we do not try this logic again.
- Add unit test for when `getXrayTraceId` throws `NoSuchMethodError`.

### DIFF
--- a/.github/actions/patch-dependencies/action.yml
+++ b/.github/actions/patch-dependencies/action.yml
@@ -103,14 +103,14 @@ runs:
       shell: bash
 
     - name: Build opentelemetry-java-instrumentation with tests
-      uses: gradle/gradle-build-action@v2
+      uses: gradle/gradle-build-action@a8f75513eafdebd8141bd1cd4e30fcd194af8dfa #v2
       if: ${{ env.patch_otel_java_instrumentation == 'true' && inputs.run_tests != 'false' }}
       with:
         arguments: check -x spotlessCheck publishToMavenLocal --scan --no-daemon
         build-root-directory: opentelemetry-java-instrumentation
 
     - name: Build opentelemetry java instrumentation
-      uses: gradle/gradle-build-action@v2
+      uses: gradle/gradle-build-action@a8f75513eafdebd8141bd1cd4e30fcd194af8dfa #v2
       if: ${{ env.patch_otel_java_instrumentation == 'true' && inputs.run_tests == 'false' }}
       with:
         arguments: publishToMavenLocal --scan --no-daemon


lambda-layer/patches/opentelemetry-java-instrumentation.patch
```diff
@@ -190,7 +192,13 @@ index 0000000000..e711558e05
 +
 +  private static String getTraceHeader(
 +      com.amazonaws.services.lambda.runtime.Context lambdaContext) {
-+    String traceHeader = lambdaContext.getXrayTraceId();
++    String traceHeader = null;
++    try {
++      // Use reflection to avoid muzzle issues with getXrayTraceId method
++      traceHeader = (String) lambdaContext.getClass().getMethod("getXrayTraceId").invoke(lambdaContext);
```
wangzlei: May cache the method "lambdaContext.getClass().getMethod("getXrayTraceId")" and reuse it for Lambda invocation later.

jj22ee: Update PR to cache method, and changed to use more efficient `MethodHandle` instead of `method`.

jj22ee:
I changed my PR to avoid Muzzle Check instead of using reflection.

lambda-layer/patches/opentelemetry-java-instrumentation.patch
```diff
@@ -162,19 +167,23 @@ index 0000000000..e711558e05
...
 + * any time.
 + */
 +public final class ParentContextExtractor {
 +
++  private static final Logger logger = Logger.getLogger(ParentContextExtractor.class.getName());
 +  private static final String AWS_TRACE_HEADER_ENV_KEY = "_X_AMZN_TRACE_ID";
 +  private static final String AWS_TRACE_HEADER_PROP = "com.amazonaws.xray.traceHeader";
 +  // lower-case map getter used for extraction
 +  static final String AWS_TRACE_HEADER_PROPAGATOR_KEY = "x-amzn-trace-id";
++  static boolean getXrayTraceIdMethodExists = true;
```
wangzlei, line 186: It is ok that only requires atomicity, does not care about visibility.

PR #1250, opened 2025-10-24
Refactor AwsMetricAttributesSpanExporter to a processor

majanjua-amzn:
## Background
With the addition of `onEnding` as part of the ExtendedSpanProcessor in OTel core (https://github.com/open-telemetry/opentelemetry-java/pull/6367), we are now able to modify spans before they are sent to an exporter. As such, we have the opportunity to refactor our `AwsMetricAttributesSpanExporter` which is responsible for adding service and dependency metric related attributes to spans as they are exported. This was done via some complicated span wrapping logic which is no longer necessary with `onEnding`.

## Changes
- Replaced all references to `AwsMetricAttributesSpanExporter` with `AwsAttributeGeneratingSpanProcessor`
- Refactored `AwsMetricAttributesSpanExporter` to `AwsAttributeGeneratingSpanProcessor`, removing any exporter-related functionality and moving the span/attribute modification logic to `processor.onEnding()`
- Replaced the `AwsMetricAttributesSpanExporterBuilder` with the `AwsAttributeGeneratingSpanProcessorBuilder`
- Updated relevant tests expecting the exporter to expect the default exporters now that we don't set this extra exporter
- Rewrote the relevant unit tests, see details in the testing section

## Testing
Converted _most_ of the exporter tests into equivalent tests for the new processor:
- Removed:
  - Tests checking testing delegate classes, export, flush, and shutdown behaviour
    - `testPassthroughDelegations`
    - `testExportDelegatingSpanDataBehaviour`
  - Tests checking >1 span being exported at a time, as the processor handles one span at a time:
    - `testExportDelegationWithMultipleSpans`
  - Helper functions that are no longer needed
- Added/renamed: (Notice some overlap, test cases before 
  - `testExportDelegationWithoutAttributeOrModification` &rarr; `testOnEndingGeneratesBothWithoutOverride`
  - `testExportDelegationWithAttributeButWithoutModification` &rarr; `testOnEndingGeneratesService`
  - `testExportDelegationWithoutAttributeButWithModification` &rarr; `testOnEndingGeneratesBothWithoutOverride`
  - `testExportDelegationWithAttributeAndModification` &rarr; `testOnEndingGeneratesBothWithOverride`
  - `testOverridenAttributes` &rarr; `testOnEndingGeneratesBothWithOverride`
  - `testExportDelegationWithTwoMetrics` &rarr; `testOnEndingGeneratesBothWithTwoMetrics`
  - `testConsumerProcessSpanHasEmptyAttribute` &rarr; `testConsumerSpanHasEmptyAttributes`
  - `testExportDelegationWithDependencyMetrics` &rarr; `testOnEndingGeneratesDependency`

Also ran the entire main build workflow using this change successfully:
https://github.com/aws-observability/aws-otel-java-instrumentation/actions/runs/18785649764


awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsApplicationSignalsCustomizerProvider.java
```diff
@@ -337,6 +337,9 @@ private SdkTracerProviderBuilder customizeTracerProviderBuilder(
       // Construct and set local and remote attributes span processor
       tracerProviderBuilder.addSpanProcessor(
           AttributePropagatingSpanProcessorBuilder.create().build());
+      // Construct and set AWS Attribute Generating Span Processor
```
thpierce, line 378: Maybe comment that the order of span processors matters. I actually wonder if it would be better to have some kind of explicit enforcement of this.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsAttributeGeneratingSpanProcessor.java
```diff
@@ -0,0 +1,95 @@
...
+
+import static software.amazon.opentelemetry.javaagent.providers.AwsAttributeKeys.AWS_SPAN_KIND;
+
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.sdk.trace.ReadWriteSpan;
+import io.opentelemetry.sdk.trace.ReadableSpan;
+import io.opentelemetry.sdk.trace.data.SpanData;
+import io.opentelemetry.sdk.trace.internal.ExtendedSpanProcessor;
```
thpierce, line 26: I forgot that `OnEnding` is in development, forcing us to depend on `internal` components explicitly not for public usage. I just want to take a moment and think about this before committing hard on this. I think fundamentally the functionality we are using is not complex and unlikely to just fundamentally change or be reverted, but [development](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/document-status.md) features are explicitly prone to that.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsAttributeGeneratingSpanProcessor.java
```diff
@@ -0,0 +1,95 @@
...
+    // Dependency metric attributes indicating the span is a local dependency root. The Service
+    // Attributes must be a subset of the Dependency, with the exception of AWS_SPAN_KIND. The
+    // knowledge that the span is a local root is more important than knowing that it is a
+    // Dependency metric, so we take all the Dependency metrics but replace AWS_SPAN_KIND with
+    // LOCAL_ROOT.
+    SpanData spanData = span.toSpanData();
+    Map<String, Attributes> attributeMap =
+        generator.generateMetricAttributeMapFromSpan(span.toSpanData(), resource);
+
+    boolean generatesServiceMetrics =
```
thpierce, line 67:
```suggestion
    SpanData spanData = span.toSpanData();
    Map<String, Attributes> attributeMap =
        generator.generateMetricAttributeMapFromSpan(spanData, resource);
    boolean generatesServiceMetrics =
```

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsAttributeGeneratingSpanProcessorTest.java
```diff
@@ -0,0 +1,258 @@
...
+            SpanKind.CLIENT.name(),
+            AttributeKey.stringKey("key"),
+            "val");
+    Map<String, Attributes> attributeMap = new HashMap<>();
+    attributeMap.put(DEPENDENCY_METRIC, metricAttributes);
+    when(generatorMock.generateMetricAttributeMapFromSpan(any(SpanData.class), any(Resource.class)))
+        .thenReturn(attributeMap);
+
+    // End span and verify it was modified appropriately
+    span.end();
```
thpierce, line 89: Curious - how does `onEnding` get called?

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsAttributeGeneratingSpanProcessorTest.java
```diff
@@ -0,0 +1,258 @@
...
+    tracer =
+        SdkTracerProvider.builder()
+            .addSpanProcessor(
+                AwsAttributeGeneratingSpanProcessor.create(generatorMock, testResource))
+            .build()
+            .get("awsxray");
+  }
+
+  @Test
+  public void testOnEndingGeneratesBothWithoutOverride() {
```
thpierce, line 71: Not blocking - a couple comments would not go amiss - e.g. why is this span a local root? What is/isn't being overridden? Both what?

TBH I find the name confusing anyways, so if not a comment, maybe a new name - same for the rest of these tests.

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsAttributeGeneratingSpanProcessorTest.java
```diff
@@ -0,0 +1,258 @@
...
+    attributeMap.put(DEPENDENCY_METRIC, metricAttributes);
+    when(generatorMock.generateMetricAttributeMapFromSpan(any(SpanData.class), any(Resource.class)))
+        .thenReturn(attributeMap);
+
+    // End span and verify it was modified appropriately
+    span.end();
+    ReadableSpan readableSpan = (ReadableSpan) span;
+    assertThat(readableSpan.getAttribute(AWS_SPAN_KIND))
+        .isEqualTo(AwsSpanProcessingUtil.LOCAL_ROOT);
+    assertThat(readableSpan.getAttribute(AWS_LOCAL_SERVICE)).isEqualTo("SERVICE");
```
thpierce, line 93:
```suggestion
    assertThat(readableSpan.getAttribute(AWS_LOCAL_SERVICE)).isEqualTo("SERVICE");
    assertThat(readableSpan.getAttribute("key")).isEqualTo("val");
```

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsAttributeGeneratingSpanProcessorTest.java
```diff
@@ -0,0 +1,258 @@
...
+    assertThat(readableSpan.getAttribute(AttributeKey.booleanKey("was_changed"))).isEqualTo(true);
```
thpierce, line 126:
```suggestion
    assertThat(readableSpan.getAttribute(AttributeKey.booleanKey("start_key"))).isEqualTo("start_val");
    assertThat(readableSpan.getAttribute(AttributeKey.booleanKey("ending_key"))).isEqualTo("ending_val");
    assertThat(readableSpan.getAttribute(AttributeKey.booleanKey("was_changed"))).isEqualTo(true);
```

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsAttributeGeneratingSpanProcessorTest.java
```diff
@@ -0,0 +1,258 @@
...
+    // 2 start attributes + 4 ending attributes - 1 overlap
+    assertThat(readableSpan.getAttributes().size()).isEqualTo(5);
+  }
+
+  @Test
+  public void testOnEndingGeneratesService() {
+    Span span = tracer.spanBuilder("test").setSpanKind(SpanKind.SERVER).startSpan();
```
thpierce, line 133: Have at least two span attribute (one replaced, one unchanged) to show we don't just completely replace attributes, etc. Same with `testOnEndingGeneratesDependency`

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsAttributeGeneratingSpanProcessorTest.java
```diff
@@ -0,0 +1,258 @@
...
+    assertThat(readableSpan.getAttribute(AWS_SPAN_KIND))
+        .isEqualTo(AwsSpanProcessingUtil.LOCAL_ROOT);
+    assertThat(readableSpan.getAttribute(AWS_LOCAL_SERVICE)).isEqualTo("SERVICE");
+    assertThat(readableSpan.getAttribute(AttributeKey.booleanKey("was_changed"))).isEqualTo(true);
+    // 2 start attributes + 4 ending attributes - 1 overlap
+    assertThat(readableSpan.getAttributes().size()).isEqualTo(5);
+  }
+
+  @Test
+  public void testOnEndingGeneratesService() {
```
thpierce, line 132:
```suggestion
  public void testOnEndingGeneratesServiceAttributes() {
```

etc.

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsAttributeGeneratingSpanProcessorTest.java
```diff
@@ -0,0 +1,258 @@
...
+          .when(
+              () ->
+                  AwsSpanProcessingUtil.shouldGenerateServiceMetricAttributes(any(SpanData.class)))
+          .thenReturn(false);
+      mockUtil
+          .when(
+              () ->
+                  AwsSpanProcessingUtil.shouldGenerateDependencyMetricAttributes(
+                      any(SpanData.class)))
+          .thenReturn(true);
```
thpierce, line 181: This is fine, but you could also probably just have a parent and child span, then audit the child span only.

awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AwsAttributeGeneratingSpanProcessorTest.java
```diff
@@ -0,0 +1,258 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.MockedStatic;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+
+/** Unit tests for {@link AwsAttributeGeneratingSpanProcessor}. */
+class AwsAttributeGeneratingSpanProcessorTest {
```
thpierce, line 50: Probably for all tests: we should include both DEPENDENCY_METRIC and SERVICE_METRIC in attributeMap, to show that only one is ever used. Actually ideally we would do something where if we expect only DEPENDENCY_METRIC to be used, then if we attempt to pull SERVICE_METRIC, we throw an exception and fail the test. Or something.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsAttributeGeneratingSpanProcessor.java
```diff
@@ -0,0 +1,95 @@
...
+
+    if (generatesServiceMetrics && generatesDependencyMetrics) {
+      // Order matters: dependency metric attributes include AWS_SPAN_KIND key
+      span.setAllAttributes(attributeMap.get(MetricAttributeGenerator.DEPENDENCY_METRIC));
+      span.setAttribute(AWS_SPAN_KIND, AwsSpanProcessingUtil.LOCAL_ROOT);
+    } else if (generatesServiceMetrics) {
+      span.setAllAttributes(attributeMap.get(MetricAttributeGenerator.SERVICE_METRIC));
+    } else if (generatesDependencyMetrics) {
+      span.setAllAttributes(attributeMap.get(MetricAttributeGenerator.DEPENDENCY_METRIC));
+    }
```
thpierce, line 80: Consideration: If (for example) `attributeMap` does NOT contain `SERVICE_METRIC` and we call `span.setAllAttributes(attributeMap.get(MetricAttributeGenerator.SERVICE_METRIC))`, what happens? If it throws an exception or something, then that is a problem, because `onEnding` is now in application path (where before export was, IIRC, async). We may need to do some defensive programming here and log debugs if expectations are not met.

PR #1294, opened 2026-01-22
Remove conflicting attributes from aws-sdk instrumentation

thpierce:
## Summary:
We are trying to upgrade to upstream's 2.23.x, but they have changed [AWS SDK 1.x/2.x  attributes to align with semantic conventions](https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v2.22.0), generally, the changes are acceptable, but observed that in our aws-sdk instrumentation package, we are duplicating attribute generation for some of these changed attributes, specifically:
* `aws.bucket.name` -> `aws.s3.bucket`
* `aws.table.name` -> `aws.dynamnodb.table_names`
* `aws.stream.name` -> `aws.kinesis.stream_name`
* `aws.queue.url` -> `aws.sqs.queue_url`

Since we would have to change these fields in the aws-sdk instrumentation, we are questioning whether we need this attribute generation logic at all - if we are just duplicating what the upstream is doing, why do it? Our aws-sdk instrumentation runs IN ADDITION TO, not IN PLACE OF the upstream, so removing this logic should be a no-op.

## Testing:
* We have [robust contract tests](https://github.com/aws-observability/aws-otel-java-instrumentation/tree/main/appsignals-tests/contract-tests/src/test/java/software/amazon/opentelemetry/appsignals/test/awssdk) that validate the `aws.*` attributes (see PR approval workflow).
* E2E tests also cover this logic, which are run in [main-build](https://github.com/aws-observability/aws-otel-java-instrumentation/actions/runs/21257574136).

### DIFF
--- a/instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsExperimentalAttributes.java
+++ b/instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsExperimentalAttributes.java
@@ -20,12 +20,8 @@ import static io.opentelemetry.api.common.AttributeKey.stringKey;
 import io.opentelemetry.api.common.AttributeKey;
 
 final class AwsExperimentalAttributes {
-  static final AttributeKey<String> AWS_BUCKET_NAME = stringKey("aws.bucket.name");
-  static final AttributeKey<String> AWS_QUEUE_URL = stringKey("aws.queue.url");
   static final AttributeKey<String> AWS_QUEUE_NAME = stringKey("aws.queue.name");
-  static final AttributeKey<String> AWS_STREAM_NAME = stringKey("aws.stream.name");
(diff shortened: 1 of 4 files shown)


instrumentation/aws-sdk/src/main/java/software/amazon/opentelemetry/javaagent/instrumentation/awssdk_v2_2/AwsExperimentalAttributes.java
```diff
@@ -20,12 +20,8 @@
 import io.opentelemetry.api.common.AttributeKey;
 
 final class AwsExperimentalAttributes {
-  static final AttributeKey<String> AWS_BUCKET_NAME = stringKey("aws.bucket.name");
-  static final AttributeKey<String> AWS_QUEUE_URL = stringKey("aws.queue.url");
   static final AttributeKey<String> AWS_QUEUE_NAME = stringKey("aws.queue.name");
```
thpierce, line 23: Callout: We can probably remove more, but limiting scope in the interest of time. Can follow up later.

PR #1355, opened 2026-04-14
feat: Add region build release workflow for Java Lambda layer

viw-test1:
*Issue #, if available:*
N/A

*Description of changes:*

Add a standalone region-build-release.yml workflow for deploying the Java Lambda layer to specific regions. This is needed for region build releases where we need to build and publish the Lambda layer independently from the full SDK release.

The workflow:

1. Takes a version and target regions as inputs (manual workflow_dispatch)
2. Builds the Java Lambda layer from source
3. Publishes the layer to each specified region in parallel
4. Generates release notes with layer ARN table, Terraform file, and CDK constants
5. Creates a draft GitHub release with the layer artifact
6. Uploads the layer artifact to the latest SDK release

This was previously part of a separate Lambda release workflow before it was merged into release-build.yml in PR #461. Re-adding it as a dedicated workflow to support region-specific deployments without triggering a full SDK release.

V2:

Key difference from previous version:
ey differences:

1. publish-layer-prod now has a Lambda layer "Upload to S3 and Sign" step that checks for a signing profile and signs the layer before publishing. The old version just uploaded and published directly.
2. publish-layer-prod uses LAYER_ARTIFACT_NAME env var instead of a hardcoded filename
3. publish-layer-prod outputs the SigningProfileVersionArn after publishing
4. generate-lambda-release-note outputs layer-note as a job output (for use by publish-github)
5. publish-github is a separate job (old version had everything in generate-release-note), and it doesn't upload to the upstream SDK release (removed that step since this is a region only build, not a full release anymore)
6. Uses the same action versions and pin hashes as the current release-build.yml

### DIFF
--- a/.github/workflows/region-build-release.yml
+++ b/.github/workflows/region-build-release.yml
@@ -1,0 +1,258 @@
+name: Region Build Release
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: The version to tag the lambda release with, e.g., 1.2.0
+        required: true
+      aws_region:
+        description: 'Deploy lambda layer to aws regions'
+        required: true
+        default: ''
+
+env:
+  LAYER_ARTIFACT_NAME: aws-opentelemetry-java-layer.zip
+  # Legacy list of commercial regions to deploy to. New regions should NOT be added here, and instead should be added to the `aws_region` default input to the workflow.
+  LEGACY_COMMERCIAL_REGIONS: us-east-1, us-east-2, us-west-1, us-west-2, ap-south-1, ap-northeast-3, ap-northeast-2, ap-southeast-1, ap-southeast-2, ap-northeast-1, ca-central-1, eu-central-1, eu-west-1, eu-west-2, eu-west-3, eu-north-1, sa-east-1
+  LAYER_NAME: AWSOpenTelemetryDistroJava
+  VERSION: ${{ github.event.inputs.version }}
+  AWS_REGION: ${{ github.event.inputs.aws_region }}
+
+permissions:
+  id-token: write
+  contents: write
+
+jobs:
+  build-layer:
+    environment: Release
+    runs-on: ubuntu-latest
+    outputs:
+      aws_regions_json: ${{ steps.set-matrix.outputs.aws_regions_json }}
+    steps:
+      - name: Set up regions matrix
+        id: set-matrix
+        env:
+          AWS_REGIONS: ${{ env.AWS_REGION }}
+        run: |
+          IFS=',' read -ra REGIONS <<< "$AWS_REGIONS"
+          MATRIX="["
+          for region in "${REGIONS[@]}"; do
+            trimmed_region=$(echo "$region" | xargs)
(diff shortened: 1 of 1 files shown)


.github/workflows/region-build-release.yml
```diff
@@ -0,0 +1,300 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: The version to tag the lambda release with, e.g., 1.2.0
+        required: true
+      aws_region:
+        description: 'Deploy lambda layer to aws regions'
+        required: true
+        default: 'us-east-1, us-east-2, us-west-1, us-west-2, ap-south-1, ap-northeast-3, ap-northeast-2, ap-southeast-1, ap-southeast-2, ap-northeast-1, ca-central-1, eu-central-1, eu-west-1, eu-west-2, eu-west-3, eu-north-1, sa-east-1, af-south-1, ap-east-1, ap-east-2, ap-south-2, ap-southeast-3, ap-southeast-4, ap-southeast-6, eu-central-2, eu-south-1, eu-south-2, il-central-1, me-central-1, me-south-1, ap-southeast-5, ap-southeast-7, mx-central-1, ca-west-1, cn-north-1, cn-northwest-1'
```
wangzlei: default is empty

.github/workflows/region-build-release.yml
```diff
@@ -0,0 +1,300 @@
...
+        run: |
+          echo "{" > ../layer_cdk
+          for file in *; do
+            read arn < "$file"
+            echo "    \"$file\": \"$arn\"," >> ../layer_cdk
+          done
+          echo "}" >> ../layer_cdk
+          cat ../layer_cdk
+
+  publish-github:
```
wangzlei: remove this step

wangzlei (changes requested):

.github/workflows/region-build-release.yml
```diff
@@ -0,0 +1,258 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: The version to tag the lambda release with, e.g., 1.2.0
+        required: true
+      aws_region:
+        description: 'Deploy lambda layer to aws regions'
+        required: true
+        default: ''
```
viw-test1, line 11: Default is empty now.

.github/workflows/region-build-release.yml
```diff
@@ -0,0 +1,258 @@
...
+      - name: generate layer ARN constants for CDK
+        working-directory: ${{ env.LAYER_NAME }}
+        run: |
+          echo "{" > ../layer_cdk
+          for file in *; do
+            read arn < "$file"
+            echo "    \"$file\": \"$arn\"," >> ../layer_cdk
+          done
+          echo "}" >> ../layer_cdk
+          cat ../layer_cdk
```
viw-test1, line 258: Publish git hub is also removed.

wangzlei (approved):

wangzlei (approved):

PR #1368, opened 2026-04-23
[Java]feat: Run auto-instrumentation init container as non-root user

Miqueasher:
## Problem
Customers enforcing `runAsNonRoot: true` via Pod Security Standards cannot use ADOT auto-instrumentation. Kubernetes checks the image metadata for a USER directive, finds none (defaults to UID 0 / root), and rejects the pod with: 
`Error: container has runAsNonRoot and image will run as root`

## Summary
Enables customers using the Kubernetes Restricted Pod Security Standard to use ADOT auto-instrumentation
Without this change, customers enforcing runAsNonRoot: true at the namespace or pod level cannot use ADOT auto-instrumentation because Kubernetes rejects the init container with:
`container has runAsNonRoot and image will run as root`

Follows the upstream OTel Operator fix (open-telemetry/opentelemetry-operator#2272)

## Changes
 Add `USER 65534:65534` to the auto-instrumentation Dockerfile so the init container runs as non-root by default

## Testing

Validated on EKS cluster (us-east-1, EKS v1.34.6) with CW Observability add-on v5.3.1

For more in-depth analysis please see [Test Results](https://quip-amazon.com/2WlNA0XUepQA/Test-Results)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,3 +52,5 @@ COPY --from=builder /usr/src/cp-utility/bin/cp-utility /bin/cp
 
 
 COPY --chmod=go+r ./otelagent/build/libs/aws-opentelemetry-agent-${ADOT_JAVA_VERSION}.jar /javaagent.jar
+
+USER 65534:65534
--- a/.github/scripts/test-adot-javaagent-image.sh
+++ b/.github/scripts/test-adot-javaagent-image.sh
@@ -6,6 +6,7 @@ TEST_TAG=$1
 ADOT_JAVA_VERSION=$2
 
 docker volume create operator-volume
+docker run --mount source=operator-volume,dst=/otel-auto-instrumentation public.ecr.aws/amazonlinux/amazonlinux:latest chmod 777 /otel-auto-instrumentation
 docker run --mount source=operator-volume,dst=/otel-auto-instrumentation ${TEST_TAG} cp /javaagent.jar /otel-auto-instrumentation/javaagent.jar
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -52,3 +52,5 @@ COPY --from=builder /usr/src/cp-utility/bin/cp-utility /bin/cp
 
 
 COPY --chmod=go+r ./otelagent/build/libs/aws-opentelemetry-agent-${ADOT_JAVA_VERSION}.jar /javaagent.jar
+
+USER 65534:65534
```
mxiamxia, line 56: Could you confirm the behavior when `USER 65534:65534` can copy the ADOT jar file during initContainer phase without permission issue? My understanding is that K8s emptyDir volumes default to `root:root 0755`, so the init container may not be able to write `javaagent.jar` to the shared volume which would affect existing customers. I'd like we can confirm this. Thanks

Miqueasher: Yes this is the confirmed behavior. It is why the docker command `docker run --mount source=operator-volume,dst=/otel-auto-instrumentation public.ecr.aws/amazonlinux/amazonlinux:latest chmod 777 /otel-auto-instrumentation` was necessary for the test to pass in the CI. Results from my local testing show that `USER 65534:65534` did copy the updated jar file with no issues. Please see the [test document](https://quip-amazon.com/2WlNA0XUepQA/Test-Results).

Miqueasher: After further testing with a clearer view inside the cluster. K8's emptyDir volumes seem to be created with `Access: (0777/drwxrwxrwx)  Uid: (    0/    root)   Gid: (    0/    root)` which is why the fix `USER 65534` worked inside of my tests.

PR #1410, opened 2026-06-26
add EC2 Dynamic Instrumentation E2E test into Application Signals E2E suite

liustve:
followup to: https://github.com/aws-observability/aws-application-signals-test-framework/pull/652

### DIFF
--- a/.github/workflows/application-signals-e2e-test.yml
+++ b/.github/workflows/application-signals-e2e-test.yml
@@ -98,6 +98,20 @@ jobs:
       java-version: '25'
       cpu-architecture: 'x86_64'
 
+  adot-di:
+    needs: [ upload-main-build ]
+    strategy:
+      fail-fast: false
+      matrix:
+        java-version: [ '8', '11', '17', '21', '25' ]
+    uses: aws-observability/aws-application-signals-test-framework/.github/workflows/java-ec2-adot-di-test.yml@main
+    secrets: inherit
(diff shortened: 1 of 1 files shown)


.github/workflows/application-signals-e2e-test.yml
```diff
@@ -98,6 +98,20 @@ jobs:
       java-version: '25'
       cpu-architecture: 'x86_64'
 
+  adot-di:
+    needs: [ upload-main-build ]
+    strategy:
+      fail-fast: false
+      matrix:
+        java-version: [ '8', '11', '17', '21', '25' ]
```
srprash, line 106: NB: How are we using these java versions as actual runtimes? Do we have sample apps built and running for each of these?

srprash (approved):

.github/workflows/application-signals-e2e-test.yml
srprash: Can you post a link here to those sample apps and how they are built?

liustve: https://github.com/aws-observability/aws-application-signals-test-framework/blob/018364c49029c7ea73f94de428f39d2716039d37/.github/workflows/java-sample-app-s3-deploy.yml#L137-L149

liustve:
image scan failing this can be bypassed as this change cannot cause such issue

PR #1423, opened 2026-07-20
feat: attribute presigned S3 URLs as AWS::S3 dependencies (opt-in, default off)

uhzzre:
## Context

Presigned URLs are a first-class Amazon S3 pattern for granting temporary, credential-free
access to a single object — commonly used for browser/mobile direct uploads and downloads and
for sharing objects across parties. The standard way to *use* a presigned URL is to issue a
plain HTTP request to it with a generic HTTP client, not through the S3 service client. AWS
documents exactly this:

- The AWS SDK for Java 2.x guide generates the URL with `S3Presigner` and then performs the
  upload/download with `java.net.HttpURLConnection`, the JDK `HttpClient`, or the SDK's
  low-level `SdkHttpClient`.
- The Amazon S3 User Guide describes uploading objects via presigned URLs with an HTTP `PUT`.

Because that HTTP request runs *below* the AWS SDK instrumentation pipeline, the S3 execution
interceptor that would normally stamp `rpc.system=aws-api` / `rpc.service=S3` never fires.

## Problem

With the call invisible to SDK instrumentation, Application Signals records a presigned S3
request as a generic HTTP call to the bucket host:

- **Attribution** — it appears as a dependency on the bucket hostname rather than an `AWS::S3`
  dependency, so it does not roll up with the service's other S3 usage on the service map.
- **Cardinality** — the remote operation becomes `<METHOD> /<object-key>`, i.e. one unique
  value per object, which inflates operation/dependency cardinality.

## What this changes

Adds an **opt-in**, generic SigV4/SigV4a presigned-URL attribution path, with S3 as the first
supported service. It complements the credential-redaction change in #1419: that PR keeps signing
material out of span attributes; this PR fixes how the call is *attributed*.

Crucially, detection is **redaction-aware** — it must work *after* the agent's default URL
sanitization has replaced the `X-Amz-Credential` and `X-Amz-Signature` values with `REDACTED`:

- The parser recognizes a presigned request from non-sensitive signals only: the presence of the
  six SigV4 query parameters (`X-Amz-Algorithm`, `X-Amz-Credential`, `X-Amz-Signature`,
  `X-Amz-Date`, `X-Amz-Expires`, `X-Amz-SignedHeaders`), with only the `X-Amz-Algorithm` value
  checked against an allowlist. Credential and signature values are never read, so a sanitized
  (redacted) URL still resolves.
- The signing service is **not** derived from the credential scope (it may be redacted); S3 is
  identified from the **endpoint hostname** — virtual-hosted and path-style forms, including
  regional, legacy regional, dual-stack, transfer acceleration, FIPS (incl. FIPS dual-stack), and
  China.
- Attributes set for a recognized S3 request:
  - `aws.remote.service = AWS::S3`
  - `aws.remote.operation` = normalized S3 operation, derived from the HTTP method, whether an
    object key is present (bucket- vs object-level), and the S3 subresource/multipart query
    parameters — covering object CRUD, `ListObjectsV2`, multipart, and the ACL / tagging /
    retention / legal-hold / torrent subresources.
  - when the bucket is identifiable: `aws.remote.resource.type = AWS::S3::Bucket`,
    `aws.remote.resource.identifier = <bucket>`
- Resolved lazily as the last fallback in remote-service/operation derivation, so it only runs
  for spans not already attributed via RPC/DB/messaging/etc., and the URL is parsed at most once
  per span.
- **Fail closed:** non-presigned URLs and non-S3 or unrecognized endpoints (custom CNAMEs, access
  points) keep the existing behavior. A recognized S3 endpoint whose bucket cannot be identified
  (e.g. a path-style request with no bucket segment) is still attributed to `AWS::S3`, without an
  `AWS::S3::Bucket` resource; its operation is normalized when identifiable and otherwise remains
  `UnknownRemoteOperation`.

## Scope

This changes only the AWS Application Signals presentation layer — the `aws.remote.*`
attribution the agent derives for the Application Signals service map and RED (Request/Error/
Duration) metrics. It does **not** modify the underlying span: the span name and the standard
HTTP/URL attributes (e.g. `url.full`, `http.request.method`) are left untouched, span names are
not rewritten, and no new spans are created. Only the values of the derived remote attributes
change (from bucket-host / `<METHOD> /<object-key>` to `AWS::S3` / normalized operation /
`AWS::S3::Bucket`). Redaction of signing material in the URL attribute is handled separately in
#1419.

## Configuration

Opt-in, disabled by default:

- `otel.aws.application.signals.presigned-url-attribution.enabled`
- `OTEL_AWS_APPLICATION_SIGNALS_PRESIGNED_URL_ATTRIBUTION_ENABLED`

Resolved from `ConfigProperties` at startup and injected into the metric attribute generator.

## Backward compatibility

Disabled by default, so existing behavior is unchanged unless the flag is set.

When enabled, attribution for presigned S3 calls changes: what previously appeared as a
dependency on the bucket host with a high-cardinality `<METHOD> /<object-key>` operation is now
attributed to `AWS::S3` with a normalized operation and, when the bucket is identifiable, an
`AWS::S3::Bucket` resource. Operators
with dashboards or alarms keyed on the previous shape should update them when enabling the flag.

## Testing

```
./scripts/local_patch.sh
./gradlew build
```

- Parameterized unit tests for the SigV4 parser (detection plus malformed / non-presigned
  rejection cases), the S3 attributor (virtual-hosted and path-style endpoint forms, operation
  mapping, and fail-closed cases), and generator precedence / remote-resource behavior.

## References

- AWS SDK for Java 2.x — Work with Amazon S3 pre-signed URLs (executes the presigned request with
  `HttpURLConnection` / JDK `HttpClient` / `SdkHttpClient`):
  https://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/examples-s3-presign.html
- Amazon S3 User Guide — Uploading objects with presigned URLs:
  https://docs.aws.amazon.com/AmazonS3/latest/userguide/PresignedUrlUploadObject.html
- Amazon S3 User Guide — Sharing objects with presigned URLs:
  https://docs.aws.amazon.com/AmazonS3/latest/userguide/ShareObjectPreSignedURL.html
- AWS General Reference — Authenticating requests: using query parameters (SigV4):
  https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html
- Related: #1419 (redact presigned URL credentials from span attributes)

## Type of change

- [x] New feature (opt-in, disabled by default)
- [x] Changes behavior when enabled: reshapes attribution of presigned S3 calls (see Backward compatibility)

## Checklist

- [x] Unit tests added/updated and passing
- [x] CHANGELOG.md updated

### DIFF
--- a/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/S3PresignedUrlAttributor.java
+++ b/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/S3PresignedUrlAttributor.java
@@ -1,0 +1,252 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.opentelemetry.javaagent.providers;
+
+import static software.amazon.opentelemetry.javaagent.providers.AwsSpanProcessingUtil.UNKNOWN_REMOTE_OPERATION;
+
+import java.util.Locale;
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import software.amazon.opentelemetry.javaagent.providers.PresignedUrlAttributor.PresignedUrlAttribution;
+import software.amazon.opentelemetry.javaagent.providers.PresignedUrlAttributor.RemoteResource;
+
+/**
+ * Derives {@code AWS::S3} attribution from a presigned S3 URL by recognizing S3 endpoint hostnames.
+ *
+ * <p>Because the signing service cannot be read from the (redacted) credential scope, S3 is
+ * identified purely from the endpoint host. Only the standard virtual-hosted and path-style S3
+ * endpoint forms are recognized. Anything else — custom CNAMEs, access points, unknown endpoints —
+ * fails closed (returns empty) so we never mis-attribute a non-S3 or unverifiable request.
+ *
+ * <p>The remote operation is derived from the HTTP method, whether an object key is present
+ * (bucket- vs object-level), and the S3 subresource/multipart query parameters. Operation names
+ * follow the S3 REST API. References:
+ *
+ * <ul>
+ *   <li>Endpoints: https://docs.aws.amazon.com/general/latest/gr/s3.html
+ *   <li>Virtual-hosted vs path-style:
+ *       https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html
+ *   <li>S3 REST API operations: https://docs.aws.amazon.com/AmazonS3/latest/API/API_Operations.html
+ * </ul>
+ */
+final class S3PresignedUrlAttributor {
+  private static final String NORMALIZED_S3_SERVICE_NAME = "AWS::S3";
+  private static final String S3_BUCKET_RESOURCE_TYPE = NORMALIZED_S3_SERVICE_NAME + "::Bucket";
+
+  // Standard S3 endpoint host forms, including global, regional, legacy regional, dual-stack,
+  // transfer acceleration, FIPS (incl. FIPS dual-stack), and China (.com.cn). The optional segment
+  // after "s3" covers the mutually exclusive endpoint styles.
+  //
+  // The legacy "-<label>" alternative is intentionally broad: besides legacy regional hosts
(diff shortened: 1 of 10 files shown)


uhzzre:
I noticed one problem here - the current implementation derives the AWS signing service from the `X-Amz-Credential` scope. This does not work with the agent’s new URL sanitization, which replaces `X-Amz-Credential`, `X-Amz-Signature`, and security-token values with `REDACTED` before metric attribution runs.

awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/S3PresignedUrlAttributor.java
```diff
@@ -0,0 +1,252 @@
...
+          && presignedAwsUrl.getFirstQueryParameterValue("torrent").isPresent()) {
+        return "GetObjectTorrent";
+      }
+    }
+
+    if (!hasObjectKey) {
+      return UNKNOWN_REMOTE_OPERATION;
+    }
+
+    switch (normalizedMethod) {
```
ezhang6811, line 202: not blocking: call out that this is a generic fallback as several unhandled S3 operations not handled above may land here (CopyObject -> PutObject, GetObjectAttributes -> GetObject).

ezhang6811 (approved):

wangzlei (approved):

PR #1429, opened 2026-07-23
fix(lambda): Cache upstream build in lambda layer build

haneric00:
*Issue #, if available:*
Lambda layer takes too long to build: we identified this is because it needs to build multiple times (4) to complete.

The two most expensive builds are to patch two OTel Java repos which only change when upstream releases a new version, and we currently clone both repos and build from scratch every time we want to complete a main build, despite this work being redundant if upstream does not change.

We propose a caching solution where we cache the final resulting build once these two upstream repos are patched, and if upstream does not change, we use this cached build instead of building and patching from step 1.

*Description of changes:*

- **Split build-layers.sh into two files**, adding build-upstream-deps.sh and build-downstream-deps.sh. For main-build and pr-build. we use a composite action that calls both build-upstream-deps.sh and build-downstream-deps.sh. If no upstream dependencies need to be built, the step is skipped and we move straight to building downstream dependencies.
  - Upstream dependencies builds using patches from OTEL, downstream dependencies builds using patches from ADOT

- **Modified workflows to include caching logic** for main build, PR build, by replacing previous build-layer.sh calls with new composite action **build-lambda-layer**
- We keep release and region build using the old build-layers.sh that forces 4 consecutive builds as a failsafe to maintain correctness

One drawback of this solution is that overtime, the github actions cache can potentially be populated with unnecessary and old builds. This can happen if upstream changes multiple times in one week. It is unlikely, but if we are to ever heavily utilize github actions caching, this has a small chance of becoming a problem. For now it is a non-issue as github actions will flush the cache periodically, and if it is ever full, the least recently used items are dropped.


.github/actions/build-lambda-layer/action.yml
```diff
@@ -0,0 +1,47 @@
...
+    - name: Restore patched upstream Maven artifacts
+      id: cache-upstream
+      uses: actions/cache/restore@0400d5f644dc74513175e3cd8d07132dd4860809 #v4.2.4
+      with:
+        path: ~/.m2/repository/io/opentelemetry/
+        # Hash every input build-upstream-deps.sh consumes. Includes
+        # .github/patches/opentelemetry-java-instrumentation.patch even though it does not exist today:
+        # the script applies it if present, so hashing it now (hashFiles ignores missing files) avoids
+        # a stale-cache hit the moment someone adds it.
+        key: ${{ runner.os }}-lambda-upstream-${{ hashFiles('.github/patches/versions', '.github/patches/opentelemetry-java-contrib.patch', '.github/patches/opentelemetry-java-instrumentation.patch', 'lambda-layer/patches/opentelemetry-java-instrumentation.patch') }}
```
jj22ee: This key hashes the inputs `build-upstream-deps.sh` *reads*, but not `build-upstream-deps.sh` itself. Edit that script — add a third upstream repo, change the `publishToMavenLocal` args, add or reorder a `patch` invocation — and every runner keeps restoring artifacts built by the old script until `versions` or a patch file happens to change. Silent, and up to a week+ long (see my other comment on this line).

Same for the toolchain: `runner.os` is in the key but the JDK isn't, and these are compiled bytecode artifacts. A `.java-version` bump (21 → next LTS) would reuse artifacts built by the old JDK.

I'd also lean on a glob rather than enumerating paths — that's what the three existing caches in this repo do (`main-build.yml:77`, `main-build.yml:234`, `pr-build.yml:190` all use `hashFiles('.github/patches/opentelemetry-java*.patch', '.github/patches/versions')`). A glob auto-covers patch files added later, which removes the need for the "hash a file that doesn't exist yet" workaround entirely.

And a manual bust segment (`-v1-`) costs nothing and saves a bad afternoon: right now, if an entry goes bad the only outs are touching a patch file (semantically wrong, and it lands in git history) or hand-deleting N entries with `gh cache delete`.

```suggestion
        key: ${{ runner.os }}-lambda-upstream-v1-${{ hashFiles('.java-version', '.github/patches/**', 'lambda-layer/patches/opentelemetry-java-instrumentation.patch', 'lambda-layer/build-upstream-deps.sh') }}
```

(Deliberately still excluding `lambda-layer/patches/aws-otel-java-instrumentation.patch` — it patches ADOT's `dependencyManagement`, which is downstream of the cache. Might be worth saying so in the comment, since someone will suggest adding it.)

jj22ee: Related but separate from the key contents: there's nothing bounding how *old* a cached upstream build can get.

The common assumption is that Actions caches age out after 7 days. They don't — that window is "since last access," and every restore resets it. `main-build.yml` restores this key on every merge to main, so the entry stays permanently warm. With a key that only moves when a human bumps `versions` or a patch, a single cached build can live for the whole lifetime of an OTel pin. We're on `2.23.0` right now, so that's realistically weeks to months.

Before this PR, every main/PR lambda build reconstructed upstream from Maven Central, so anything not perfectly pinned got refreshed continuously. Now it's frozen until someone bumps a version.

A rotating segment caps the worst case at 7 days regardless of activity:

```yaml
- id: cache-window
  shell: bash
  run: echo "stamp=$(date -u +%G-%V)" >> "$GITHUB_OUTPUT"   # ISO year-week
```

then fold `${{ steps.cache-window.outputs.stamp }}` into the key. Cost is one full build per week — roughly what we were paying on *every* build before this PR.

Context for why I'm picking at this: `nightly-upstream-snapshot-build.yml:41` keys on `hashFiles('.github/patches/opentelemetry-java*.patch')` and omits `.github/patches/versions` entirely — in the one workflow whose whole job is validating against upstream **SNAPSHOTs**, which are mutable by definition. That cache can pin a weeks-old snapshot while the job reports it tested "latest upstream." Not this PR's bug (and this PR is already better, since it does hash `versions`), but it's the same shape and I'd rather we not add a second instance. Happy to file that one separately.

.github/actions/build-lambda-layer/action.yml
```diff
@@ -0,0 +1,47 @@
...
+  first and the result is cached for future runs.
+
+runs:
+  using: "composite"
+  steps:
+    - name: Restore patched upstream Maven artifacts
+      id: cache-upstream
+      uses: actions/cache/restore@0400d5f644dc74513175e3cd8d07132dd4860809 #v4.2.4
+      with:
+        path: ~/.m2/repository/io/opentelemetry/
```
jj22ee: Heads up that three jobs already cache this exact directory under a different prefix: `main-build.yml:77`, `main-build.yml:234`, `pr-build.yml:190`, all `${{ runner.os }}-maven-local-...`.

Those are populated by `patch-dependencies`, which never applies `lambda-layer/patches/opentelemetry-java-instrumentation.patch`. So we now have two caches holding **different bytes at identical Maven coordinates**, separated only by key prefix. That's safe today, and the absence of `restore-keys` here is exactly right — but it's load-bearing in a way that isn't obvious. If anyone ever adds a prefix fallback to either cache, or merges these jobs, we'd silently ship jars built from the wrong patch set.

Could you add a comment saying the prefix separation is intentional and `restore-keys` must not be added? It's the kind of thing a well-meaning reviewer "fixes" later.

.github/actions/build-lambda-layer/action.yml
```diff
@@ -0,0 +1,47 @@
...
+
+  The upstream OpenTelemetry contrib and instrumentation builds are cached in GitHub Actions
+  cache, keyed on the pinned versions and patch files. On a cache hit, only the ADOT build
+  (from current source) and the layer packaging run. On a miss, the full upstream build runs
+  first and the result is cached for future runs.
+
+runs:
+  using: "composite"
+  steps:
+    - name: Restore patched upstream Maven artifacts
```
jj22ee: On a subtly incomplete or wrong-content hit, the symptom is a Gradle `Could not find io.opentelemetry.javaagent:...:2.23.0-adot-lambda1` buried in the downstream build, with no hint that the cache caused it and no documented way to bust it.

A cheap sanity step right after restore would help a lot: read the versions out of `.github/patches/versions` and assert the expected artifacts are actually present under `~/.m2/repository/io/opentelemetry/`; if not, clear `cache-hit` and fall through to the full build. That also closes the partial-cache hole more robustly than the `success()` guard, since it catches a bad entry on *read* rather than trying to prevent every bad *write*.

.github/actions/build-lambda-layer/action.yml
```diff
@@ -0,0 +1,47 @@
+name: "Build Lambda Layer"
+description: |
```
jj22ee: The action assumes the caller has already checked out the repo, set up Java per `.java-version`, and is running with the workspace root as the working directory. Worth stating in `description:`, the way `patch-dependencies` does ("This action assumes that java was set correctly") — it's an implicit contract that's easy to break from a new caller.

Also worth noting somewhere: composite action steps can't use `continue-on-error`, so a cache service hiccup on restore or save now fails the whole lambda job. That job previously had no dependency on the Actions cache service. Probably fine, but it's a new failure mode — and worth confirming that a duplicate-key reservation race (main-build plus a few PRs all missing at once) only warns rather than fails the step.

lambda-layer/build-upstream-deps.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+# Builds the PATCHED upstream OpenTelemetry dependencies (contrib + instrumentation)
+# These are the slow, rarely-changing steps of the lambda layer build. They only
+# need to rerun when the pinned upstream versions (.github/patches/versions) or the
+# patch files change. In CI, this script is skipped on an actions/cache hit keyed on
+# exactly those files; build-layer.sh (ADOT from source + package) always runs after.
```
jj22ee: `build-layer.sh` isn't what runs after this — it's `build-downstream-deps.sh`. `build-layer.sh` is the full monolith that runs *instead of* both. Worth fixing since it's a brand-new file and this line is the first thing someone reads when they touch the caching logic.

```suggestion
# exactly those files; build-downstream-deps.sh (ADOT from source + package) always runs after.
```

lambda-layer/build-downstream-deps.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+# Builds ADOT Java from the current source and packages the Lambda layer artifact.
+# This is the "downstream" half of the layer build: it assumes the patched upstream
+# OpenTelemetry dependencies (contrib + instrumentation) produced by build-upstream-deps.sh
+# are already published to the local Maven repo (~/.m2), either from a fresh
+# build-upstream-deps.sh run or an actions/cache restore. It always runs in CI.
+#
+# For a full, from-scratch build (upstream + downstream in one shot) — e.g. releases where we
+# want to guarantee no cached-state mistakes — use build-layer.sh instead.
```
jj22ee: `build-upstream-deps.sh` + `build-downstream-deps.sh` concatenated is `build-layer.sh` verbatim, plus a duplicated `awk` version-extraction block. With release on the monolith and PR/main on the split pair, we're validating a different code path than we release from, and every future fix has to be applied in two places or they drift.

Nearly free to fix — replace the body of `build-layer.sh` with:

```bash
set -e
SOURCEDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
"$SOURCEDIR"/build-upstream-deps.sh
"$SOURCEDIR"/build-downstream-deps.sh
```

Zero behavior change for release, one source of truth, and the duplicated version block goes away.

Separately, minor: `build-downstream-deps.sh` is a slightly misleading name — it builds ADOT and packages the layer, it doesn't build dependencies. `build-layer-only.sh` or `package-layer.sh` reads better, and these names will outlive the PR.

lambda-layer/build-downstream-deps.sh
```diff
@@ -0,0 +1,50 @@
...
+echo "Info: Building ADOT Java from current source"
+pushd "$SOURCEDIR"/..
+patch  -p1 < "${SOURCEDIR}"/patches/aws-otel-java-instrumentation.patch
+CI=false ./gradlew publishToMavenLocal -Prelease.version=${version}-adot-lambda1
+popd
+
+
+## Build ADOT Lambda Java SDK Layer Code
+echo "Info: Building ADOT Lambda Java SDK Layer Code"
+./gradlew build -PotelVersion=${otel_instrumentation_version} -Pversion=${version}
```
jj22ee: Not introduced here — `build-layer.sh` has the same line — so not asking you to fix it in this PR. Flagging it because the split means it now has to be fixed in two places (or once, if you take the wrapper suggestion above).

It expands to empty, so this is effectively `-PotelVersion=`. Harmless today only because `lambda-layer/build.gradle.kts` reads `otelVersion` in `compileOnly` configurations that never get resolved. It looks like it was meant to be `$version`.

.github/actions/build-lambda-layer/action.yml
jj22ee: This key hashes what `build-upstream-deps.sh` reads, but not the script itself. Change the script — a new upstream repo, different `publishToMavenLocal` args, a reordered `patch` call — and every runner keeps restoring artifacts built by the old script until `versions` or a patch file happens to change. `.java-version` has the same gap: `runner.os` is in the key but the JDK isn't, and these are compiled bytecode artifacts.

Suggest hashing the script and the JDK, switching to a glob (matches the three existing caches in this repo and removes the need for the "hash a file that doesn't exist yet" workaround), and adding a `-v1-` segment so a bad entry can be busted without touching a patch file.

```suggestion
        key: ${{ runner.os }}-lambda-upstream-v1-${{ hashFiles('.java-version', '.github/patches/**', 'lambda-layer/patches/opentelemetry-java-instrumentation.patch', 'lambda-layer/build-upstream-deps.sh') }}
```

lambda-layer/build-downstream-deps.sh
jj22ee: `build-upstream-deps.sh` and `build-downstream-deps.sh` concatenated are `build-layer.sh` verbatim, plus a duplicated version-extraction block. Since release uses the monolith and main/PR use the split pair, we validate a different path than we release from, and every future fix has to land twice. I see 25158bc did split it and df3273b restored it to keep release from-scratch — right goal, but the copy is the cost.

A wrapper gets the same property with no duplication: identical execution for release, no cache involvement, and `build-layer.sh` keeps working unchanged for local Mac builds.

```bash
set -e
SOURCEDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
"$SOURCEDIR"/build-upstream-deps.sh
"$SOURCEDIR"/build-downstream-deps.sh
```

.github/actions/build-lambda-layer/action.yml
```diff
@@ -0,0 +1,47 @@
...
+      uses: actions/cache/restore@0400d5f644dc74513175e3cd8d07132dd4860809 #v4.2.4
+      with:
+        path: ~/.m2/repository/io/opentelemetry/
+        # Hash every input build-upstream-deps.sh consumes. Includes
+        # .github/patches/opentelemetry-java-instrumentation.patch even though it does not exist today:
+        # the script applies it if present, so hashing it now (hashFiles ignores missing files) avoids
+        # a stale-cache hit the moment someone adds it.
+        key: ${{ runner.os }}-lambda-upstream-${{ hashFiles('.github/patches/versions', '.github/patches/opentelemetry-java-contrib.patch', '.github/patches/opentelemetry-java-instrumentation.patch', 'lambda-layer/patches/opentelemetry-java-instrumentation.patch') }}
+
+    - name: Build patched upstream dependencies
```
jj22ee: Every PR used to prove the whole chain works: clone, checkout tag, apply patches, build. With this step skipped on a hit, most PRs only exercise the downstream half — so a break in the upstream half (upstream retagging, a patch going stale, plugin resolution shifting) surfaces on the next version-bump PR or at release. Release is the worst place to find it, and it runs `build-layer.sh`, which nothing in CI touches.

A scheduled nightly or weekly job running `build-layer.sh` end-to-end restores that signal for one build's worth of runner time.

.github/actions/build-lambda-layer/action.yml
jj22ee: Nothing bounds how old an entry can get. The 7-day window is since-last-access, not age, and `main-build.yml` restores this key on every merge to main, so the clock resets continuously and the entry never expires. With a key that only moves on a human version bump, one cached upstream build can live the entire lifetime of an OTel pin — weeks to months on `2.23.0`.

A rotating segment folded into the key caps the worst case at 7 days regardless of activity, for one extra full build per week.

```yaml
- id: cache-window
  shell: bash
  run: echo "stamp=$(date -u +%G-%V)" >> "$GITHUB_OUTPUT"   # ISO year-week
```

lambda-layer/build-downstream-deps.sh
```diff
@@ -0,0 +1,50 @@
...
+# Exit if the version is empty or null
+if [[ -z "$version" ]]; then
+  echo "Error: Version could not be found in ${file}."
+  exit 1
+fi
+
+## Build the ADOT Java from current source
+echo "Info: Building ADOT Java from current source"
+pushd "$SOURCEDIR"/..
+patch  -p1 < "${SOURCEDIR}"/patches/aws-otel-java-instrumentation.patch
```
jj22ee: `set -e` aborts before the `git restore` on line 50, so any mid-build failure leaves this patch applied. The next run's `patch -p1` then hits "Reversed (or previously applied) patch detected!  Assume -R?" and prompts on the tty. CI never sees it because every job is a fresh checkout, but it will bite local iteration — and the split makes local iteration attractive, since re-running just this script is now the fast inner loop.

A trap makes the revert unconditional:

```bash
trap 'git -C "$SOURCEDIR/.." restore dependencyManagement/build.gradle.kts' EXIT
```

.github/actions/build-lambda-layer/action.yml
jj22ee: On an incomplete or wrong-content hit, the symptom is a Gradle `Could not find io.opentelemetry.javaagent:...:2.23.0-adot-lambda1` deep in the downstream build, with nothing pointing at the cache as the cause and no documented way to bust it.

A step after restore that reads `.github/patches/versions` and asserts the expected artifacts exist under `~/.m2/repository/io/opentelemetry/`, falling through to the full build if not, catches a bad entry on read instead of trying to prevent every bad write.

.github/actions/build-lambda-layer/action.yml
jj22ee: Three jobs already cache this exact directory under `${{ runner.os }}-maven-local-...` (`main-build.yml:77`, `main-build.yml:234`, `pr-build.yml:190`). Those are built by `patch-dependencies`, which never applies `lambda-layer/patches/opentelemetry-java-instrumentation.patch` — so two caches now hold different bytes at identical Maven coordinates, separated only by key prefix. Adding a prefix fallback to either, or merging these jobs, would silently ship jars built from the wrong patch set.

It's correct as written with no `restore-keys`; a comment saying the prefix separation is intentional and `restore-keys` must not be added would keep it that way.

lambda-layer/build-upstream-deps.sh
jj22ee: It's `build-downstream-deps.sh` that runs after this, not `build-layer.sh` — `build-layer.sh` replaces both halves.

```suggestion
# exactly those files; build-downstream-deps.sh (ADOT from source + package) always runs after.
```

wangzlei (approved):

.github/actions/build-lambda-layer/action.yml
jj22ee, line 22: Can we fold `build-upstream-deps.sh` itself, and `.java-version`, into this hash? Right now the key covers the files that script *reads*, but not the script, and not the JDK.

That becomes a problem the next time someone edits it. Say we find a bug in how the contrib patch gets applied and fix it — `versions` and the patch files are untouched, so the key is identical, CI reports a cache hit, and the script never runs. The build goes green and the fix looks verified, but the layer is still built from jars the old script produced, and it stays that way until someone unrelated bumps an OTel version. The JDK gap is the same shape: these are compiled bytecode artifacts, and `runner.os` is in the key but the Java version isn't.

Switching to a glob would also match the three existing caches in this repo and remove the need for the "hash a file that doesn't exist yet" workaround, and a `-v1-` segment gives us a way to bust a bad entry without touching a patch file.

```yaml
key: ${{ runner.os }}-lambda-upstream-v1-${{ hashFiles('.java-version', '.github/patches/**', 'lambda-layer/patches/opentelemetry-java-instrumentation.patch', 'lambda-layer/build-upstream-deps.sh') }}
```

lambda-layer/build-downstream-deps.sh
jj22ee, line 9: Could `build-layer.sh` call these two scripts instead of carrying its own copy of them? It isn't part of this PR, which is what makes this easy to miss: `build-upstream-deps.sh` and `build-downstream-deps.sh` concatenated are `build-layer.sh` verbatim, so the full build now exists in two places — and the copy release uses is the one no CI job runs.

The cost is already visible right here. Line 35 passes `-PotelVersion=${otel_instrumentation_version}`, which is undefined and expands to empty. The identical line sits in `build-layer.sh`, so fixing it in this file wouldn't fix what release actually executes, and nothing would flag the gap.

This keeps release exactly as it is today — same execution order, no cache involvement — and `build-layer.sh` still works unchanged for local builds.

```bash
# build-layer.sh
set -e
SOURCEDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
"$SOURCEDIR"/build-upstream-deps.sh
"$SOURCEDIR"/build-downstream-deps.sh
```

jj22ee (commented):
A couple comments, these can be addressed in a follow-pr since this was already merged.

PR #1436, opened 2026-08-07
Add CloudWatch Plugin for OpenTelemetry (span metrics)

srprash:
## Description

Adds a new standalone module, **`cloudwatch-plugin-otel`**, a plugin for the OpenTelemetry Java SDK
that generates request ("RED") metrics from spans **in-process, from 100% of spans, before trace
sampling** — solving the undercount that occurs when span metrics are generated downstream from only
the sampled subset of spans.

Span metrics are the first feature of this plugin.

### What it produces

- `traces.span.metrics.calls` (counter) and `traces.span.metrics.duration` (histogram, seconds),
  emitted under the instrumentation scope `cloudwatch.plugin.otel.span_metrics`, riding the host
  SDK's `MeterProvider`.
- A record-forcing sampler (`DROP` -> `RECORD_ONLY`) so metrics see every span while trace export
  still honors the configured sampling rate.
- Low-cardinality dimensions: base attributes plus allowlisted HTTP/RPC/DB/messaging semconv
  attributes present on the span (current keys, with legacy semconv keys passed through as a
  fallback).

### Wiring modes (zero plugin configuration)

Java agent extension, Spring Boot starter, plain SDK autoconfigure, and manual SDK.

### Coordinate

`software.amazon.opentelemetry:cloudwatch-plugin-otel`

### Compatibility

Verified against OpenTelemetry Java SDK 1.32.0 or later (built/tested on 1.44.x). Core targets
Java 8 (compiled at `release 8` and exercised by a Java 8 runtime smoke test); the Spring Boot
starter mode requires Java 17 / Spring Boot 3. The plugin bundles no OpenTelemetry.

## Testing

- Unit tests (sampler incl. trace-state forwarding, processor emission, attribute building incl.
  current/legacy semconv precedence, lifecycle/exception-safety, no-metrics-pipeline inertness, SPI).
- Unit tests run at the minimum supported SDK (1.32.0) and the latest release; a nightly workflow
  runs the contract tests for all four wiring modes at floor/current/latest.
- Contract tests across all four wiring modes and per-family attribute shape (HTTP, DB, RPC,
  messaging) using Testcontainers + a mock collector.
- Java 8 runtime smoke test on a JDK 8 toolchain.

## Follow-ups (separate PRs, after this merges)

- Maven Central release/publish workflow.
- README/docs polish (calls-unit table cell, documenting the schema/lib-version marker attributes
  and the span attribute mutation, and noting configurations the plugin does not wire into).

Remaining review nits on this PR are non-blocking and tracked to be addressed in follow-up PRs.


cloudwatch-plugin-otel/CHANGELOG.md
```diff
@@ -0,0 +1,19 @@
+# Changelog
```
liustve, line 1: TODO: 
Let's see if we can integrate this into our existing CHANGELOG.md and reuse the same workflow

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/internal/SpanMetricsCustomizerProvider.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public final class SpanMetricsCustomizerProvider implements AutoConfigurationCustomizerProvider {
+
+  @Override
+  public void customize(AutoConfigurationCustomizer autoConfiguration) {
+    autoConfiguration.addSamplerCustomizer(
+        (sampler, config) -> AlwaysRecordSampler.create(sampler));
+    autoConfiguration.addTracerProviderCustomizer(
+        (tracerProvider, config) -> tracerProvider.addSpanProcessor(new SpanMetricsProcessor()));
+  }
```
liustve, line 45: Mode 3 (plain autoconfigure) looks broken here: `customize` wraps the sampler and registers the processor but never sets `OpenTelemetryHolder`, so `ensureInstruments` bails on every `onEnd` and no metrics come out. Simplest fix is to have this provider also implement `AutoConfigureListener` and set the holder in `afterAutoConfigure`.

cloudwatch-plugin-otel/contract-tests/apps/autoconfigure-app/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/e2e/app/AutoconfigureApp.java
```diff
@@ -0,0 +1,177 @@
...
+
+  private static Tracer tracer;
+
+  private AutoconfigureApp() {}
+
+  public static void main(String[] args) throws Exception {
+    OpenTelemetrySdk sdk =
+        AutoConfiguredOpenTelemetrySdk.builder().setResultAsGlobal().build().getOpenTelemetrySdk();
+    // Manual autoconfigure setups must bind the built SDK so the span processor can obtain a Meter.
+    SpanMetrics.bind(sdk);
```
liustve: Test gap: `AutoconfigureModeTest` passes only because this line calls `SpanMetrics.bind(sdk)` itself, which is really mode 4 (manual) wiring. So the actual "no code changes" autoconfigure path a real user hits is never exercised. Dropping this manual bind would make the test cover the real path (and catch the mode 3 issue).

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/AlwaysRecordSampler.java
```diff
@@ -0,0 +1,80 @@
...
+      Context parentContext,
+      String traceId,
+      String name,
+      SpanKind spanKind,
+      Attributes attributes,
+      List<LinkData> parentLinks) {
+    SamplingResult result =
+        delegate.shouldSample(parentContext, traceId, name, spanKind, attributes, parentLinks);
+    if (result.getDecision() == SamplingDecision.DROP) {
+      return SamplingResult.create(SamplingDecision.RECORD_ONLY, result.getAttributes());
```
liustve: This drops trace-state on the DROP path: `SamplingResult.create(RECORD_ONLY, ...)` uses the default `getUpdatedTraceState`, which returns the parent unchanged, so anything the delegate set is lost. All three reference implementations forward it instead, wrapping the result and delegating `getUpdatedTraceState`:

- [OTel incubator sampler](https://github.com/open-telemetry/opentelemetry-java/blob/8e332fc34abf6d62ae3d23eedf405211382e4a71/sdk-extensions/incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/samplers/AlwaysRecordSampler.java#L91-L92) (the one this comment cites)
- [contrib aws-xray](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/fbf83046883640b2814ce9a3587af734e8319702/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AlwaysRecordSampler.java#L80)
- [existing `awsagentprovider` sampler in this repo](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/e3ccbe584b78c6692a7bf8a3d9a9d5b495d3b529/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AlwaysRecordSampler.java#L89-L90)

Harmless under the current parentBased/traceIdRatio samplers, but it deviates from all of them despite the comment saying there's nothing to keep in sync.

cloudwatch-plugin-otel/src/test/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/AlwaysRecordSamplerTest.java
```diff
@@ -0,0 +1,79 @@
...
+            Collections.emptyList());
+  }
+
+  @Test
+  void dropBecomesRecordOnly() {
+    Attributes attrs = Attributes.of(AttributeKey.stringKey("k"), "v");
+    SamplingResult result = sample(SamplingDecision.DROP, attrs);
+    assertThat(result.getDecision()).isEqualTo(SamplingDecision.RECORD_ONLY);
+    assertThat(result.getAttributes()).isEqualTo(attrs);
+  }
```
liustve, line 65: Test gap: `dropBecomesRecordOnly` only checks `getDecision` and `getAttributes`, so the trace-state drop slips through. The [`awsagentprovider` `AlwaysRecordSamplerTest`](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/e3ccbe584b78c6692a7bf8a3d9a9d5b495d3b529/awsagentprovider/src/test/java/software/amazon/opentelemetry/javaagent/providers/AlwaysRecordSamplerTest.java#L95-L97) asserts `getUpdatedTraceState` matches the delegate (with a root result that actually mutates it); porting that one assertion here would catch it.

cloudwatch-plugin-otel/README.md
```diff
@@ -0,0 +1,115 @@
...
+`status.code`, plus any allowlisted semantic-convention attributes present on the span
+(e.g. `http.request.method`, `http.route`, `http.response.status_code`, `rpc.system`/`service`/`method`,
+`db.system.name`/`operation.name`/`collection.name`, `messaging.system`/`operation.name`/`destination.name`).
+
+The metrics are emitted under the instrumentation scope `otel.cloudwatch.spanmetrics` and ride your
+application's existing `MeterProvider` — they flow wherever your other metrics already go.
+
+## Requirements & compatibility
+
+- **OpenTelemetry Java SDK 1.32.0 – 1.64.0** (verified). Built and tested against 1.44.x.
```
liustve: The tests run against a single SDK version, so this range isn't exercised at either bound. Unit tests resolve to 1.44.1 (via the instrumentation BOM) and the contract-test apps to 1.44/1.45, so nothing runs at the 1.32.0 floor or the 1.64.0 ceiling that "verified" covers. OTel tests its instrumentation libraries against [both the oldest and the latest supported version](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/eeac17eea038953ae2053d61df4a78e4fb72a90b/docs/contributing/running-tests.md), the default run pinned to the minimum and testLatestDeps re-running the same suite against the newest.

cloudwatch-plugin-otel/src/test/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/SpanMetricsProcessorTest.java
```diff
@@ -0,0 +1,166 @@
...
+  @BeforeAll
+  static void setUp() {
+    // The holder is a first-wins singleton, so all tests share one bound SDK.
+    reader = InMemoryMetricReader.create();
+    SdkMeterProvider meterProvider =
+        SdkMeterProvider.builder().registerMetricReader(reader).build();
+    OpenTelemetrySdk sdk = OpenTelemetrySdk.builder().setMeterProvider(meterProvider).build();
+    SpanMetrics.bind(sdk);
+    processor = new SpanMetricsProcessor();
+  }
```
liustve, line 54: Every test binds an SDK here before anything runs, so `ensureInstruments` always finds a holder and the null branch in `onEnd` never executes. That branch (`OpenTelemetryHolder.get() == null` returns false and silently drops the metric) is the exact production state of plain autoconfigure mode, where nothing binds. The holder is a first-wins static singleton, so once this `@BeforeAll` sets it the unbound path is unreachable for the rest of the JVM, which is why a test would need its own classloader or a reset hook to reach it.

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/internal/SpanMetricsCustomizerProvider.java
srprash: Fixed in 8356ee2: SpanMetricsCustomizerProvider now also implements AutoConfigureListener and binds the built SDK in afterAutoConfigure, so plain autoconfigure emits metrics without a manual bind.

cloudwatch-plugin-otel/contract-tests/apps/autoconfigure-app/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/e2e/app/AutoconfigureApp.java
srprash: Fixed in 8356ee2: dropped the manual SpanMetrics.bind from AutoconfigureApp so the test now exercises the real no-code autoconfigure path.

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/AlwaysRecordSampler.java
srprash: Fixed in 8356ee2: the DROP path now returns a wrapper that forwards the delegate's getUpdatedTraceState instead of discarding it, matching the reference samplers.

cloudwatch-plugin-otel/src/test/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/AlwaysRecordSamplerTest.java
srprash: Fixed in 8356ee2: added dropForwardsDelegateTraceState, which asserts the RECORD_ONLY result's getUpdatedTraceState matches a delegate that mutates trace-state.

cloudwatch-plugin-otel/README.md
srprash: Addressed in a8ec0ea/b42b170: PR build now runs unit tests at 1.32.0 and against latest.release, and a nightly workflow runs the SDK-facing contract tests against the latest SDK. README reworded to '1.32.0 or later'.

cloudwatch-plugin-otel/src/test/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/SpanMetricsProcessorTest.java
srprash: Fixed in 292d5b1: added a test-only OpenTelemetryHolder.reset() and OpenTelemetryHolderTest, including processorDropsMetricsWhenHolderUnbound which exercises the null-holder branch (onEnd drops without throwing).

.github/workflows/cloudwatch-plugin-otel-nightly-compat.yml
```diff
@@ -0,0 +1,36 @@
...
+          distribution: temurin
+
+      - name: Build extension jar
+        run: ./gradlew jar -p cloudwatch-plugin-otel --stacktrace
+
+      - name: Run SDK-facing contract tests against latest OTel SDK
+        run: |
+          ./gradlew contractTests -p cloudwatch-plugin-otel/contract-tests \
+            -PotelBomVersion=latest.release \
+            --tests "*ManualModeTest" --tests "*AutoconfigureModeTest" \
```
liustve: Test gap: minimum/latest behavioral coverage is still partial. The PR runs `1.32.0` and latest for unit tests, but contracts run once at pinned versions; nightly latest selects only manual and autoconfigure. Javaagent remains pinned to `2.29.0`, Spring to BOM `2.10.0`, and no contracts run at the SDK floor. We do not need every release, but each supported wiring surface should cover minimum/current/latest using separate version inputs where necessary.

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/internal/OpenTelemetryHolder.java
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * <p>This class is internal and not part of the public API.
+ */
+public final class OpenTelemetryHolder {
+
+  private static final Logger logger = Logger.getLogger(OpenTelemetryHolder.class.getName());
+
+  private static final AtomicReference<OpenTelemetry> instance = new AtomicReference<>();
+
+  /** First bind wins. A later bind of a different instance is ignored but logged. */
```
liustve, line 39: nit: `OpenTelemetryHolder` is static and first-bind-wins per classloader. If a second Spring context or autoconfig SDK is created, its processor still records through the first SDK's `MeterProvider`, potentially using the wrong exporter and metric resource. Can we make the binding per customizer/processor instance, or explicitly document one SDK per classloader?

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/internal/SpanMetricsCustomizerProvider.java
```diff
@@ -0,0 +1,47 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.opentelemetry.cloudwatch.spanmetrics.internal;
+
+import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.sdk.autoconfigure.spi.AutoConfigurationCustomizer;
+import io.opentelemetry.sdk.autoconfigure.spi.AutoConfigurationCustomizerProvider;
+import io.opentelemetry.sdk.autoconfigure.spi.internal.AutoConfigureListener;
```
liustve, line 21: nit: This depends on `io.opentelemetry.sdk.autoconfigure.spi.internal.AutoConfigureListener`. OTel excludes `.internal` packages from compatibility guarantees, but this hook is central to the advertised `1.32.0+` support. Since there is no equivalent stable post-build hook, can we document the dependency and ensure minimum/latest tests actually load this SPI path?

cloudwatch-plugin-otel/build.gradle.kts
```diff
@@ -0,0 +1,222 @@
...
+      }
+    }
+  }
+}
+
+java {
+  withSourcesJar()
+  withJavadocJar()
+  sourceCompatibility = JavaVersion.VERSION_1_8
+  targetCompatibility = JavaVersion.VERSION_1_8
```
liustve, line 116: nit: The module advertises Java 8 support, but `sourceCompatibility`/`targetCompatibility` on JDK 21 do not prevent compiling against Java 9+ APIs. Current tests run on JDK 21 and contract containers use Java 23. Can we use `options.release.set(8)` for main sources and add a minimal Java 8 runtime smoke test?

cloudwatch-plugin-otel/settings.gradle.kts
```diff
@@ -0,0 +1,19 @@
+rootProject.name = "cloudwatch-plugin-otel"
```
liustve, line 1: nit: This is a standalone Gradle build and is not included by the root `settings.gradle.kts`, so the repository's Spotless and JaCoCo conventions do not apply. The dedicated workflow's `build -p cloudwatch-plugin-otel` therefore misses those gates. Can we apply the same formatting/license and coverage conventions inside this build and run `spotlessCheck` in its workflow?

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/internal/SpanMetricsCustomizerProvider.java
srprash: done. added clarifying comment and covered in test.

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/internal/OpenTelemetryHolder.java
srprash: Multiple SDKs in a single JVM is pretty rare in practice and an edge case. To address would mean a non-trivial refactor. I suggest we skip and only address if an issue is reported.

.github/workflows/cloudwatch-plugin-otel-nightly-compat.yml
srprash: done

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/SpanMetricsProcessor.java
```diff
@@ -0,0 +1,148 @@
...
+  private volatile LongCounter calls;
+  private volatile DoubleHistogram duration;
+
+  @Override
+  public void onStart(Context parentContext, ReadWriteSpan span) {
+    // Dedup + schema/lib markers: presence of the schema attribute tells the backend this span's
+    // metrics were already generated by the extension (do not regenerate). Guarded so a failure
+    // here can never propagate into the host application's span pipeline.
+    try {
+      span.setAttribute(SCHEMA_ATTR, SCHEMA_VERSION);
```
liustve, line 99: The backend-suppression marker is written before the plugin knows that any client metric can be delivered. With `OTEL_METRICS_EXPORTER=none` or a reader-less `MeterProvider`, the SDK supplies no-op instruments, `ensureInstruments()` succeeds, and exported spans still carry `v1`; the v186 dedup rule then makes the backend skip generation, so no side produces metrics. A real autoconfigure probe reproduced this for every span. Please resolve the exact-once invariant for explicit-none/reader-less configurations, document any unsupported state, and add a real-span regression with no metric reader.

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/SpanMetricsAttributesBuilder.java
```diff
@@ -0,0 +1,141 @@
...
+  static final AttributeKey<String> STATUS_CODE = AttributeKey.stringKey("status.code");
+
+  // Copied when present; a span only carries the keys of its own family, so no family branching.
+  private static final List<AttributeKey<?>> ALLOWLIST =
+      Arrays.asList(
+          AttributeKey.stringKey("http.request.method"),
+          AttributeKey.longKey("http.response.status_code"),
+          AttributeKey.stringKey("http.route"),
+          AttributeKey.stringKey("error.type"),
+          AttributeKey.stringKey("rpc.system"),
```
liustve: The v186 table defines `rpc.system.name` as the current key and `rpc.system` only as its legacy fallback, but this unconditional allowlist does the inverse: current-only spans lose the RPC-system dimension, and dual-emission spans retain the legacy key. Please allowlist `rpc.system.name`, move `rpc.system` into a current-first fallback pair, and cover current-only, legacy-only, and both-key precedence. The existing legacy-only assertions should remain valid.

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/SpanMetricsProcessor.java
srprash: Fixed in 59216df. When the bound `MeterProvider` has no reader (e.g. `OTEL_METRICS_EXPORTER=none` or a reader-less provider), the SDK returns the API no-op meter, so recording goes nowhere while the marker would still suppress downstream generation. The processor now detects the no-op meter and stays inert in that case — it does not stamp the marker and does not record — so the downstream generator still produces the metric from the sampled spans and the exactly-once behavior holds. The check is on both the `onStart` stamp and the `onEnd` recording so they can't disagree.

Detection compares the returned meter's class against `MeterProvider.noop().get(...)`'s meter class rather than a hard-coded class name, since the concrete no-op class differs across SDK versions.

Added `NoMetricsPipelineTest` covering the reader-less case (no marker, no throw) and the with-reader control (marker + `calls` recorded). It runs under the min/latest unit-test matrix, so a future SDK change to reader-less `MeterProvider` behavior that breaks the detection fails the build.

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/SpanMetricsAttributesBuilder.java
srprash: Fixed in 90191be. The allowlist now uses the current key `rpc.system.name`, with `rpc.system` moved into a current-first legacy fallback (passed through under its own key/value when the current key is absent) — matching the current OTel semantic conventions, where `rpc.system` is deprecated in favor of `rpc.system.name`.

While here I also aligned the HTTP family the same way (current `http.request.method` / `http.response.status_code` with legacy fallbacks `http.method` / `http.status_code`); the legacy-fallback helper was generalized so the long-valued `http.status_code` passes through unchanged. Database fallbacks were already current-first.

Covered with current-only, legacy-only, and both-present (precedence) unit tests for the RPC and HTTP pairs; the existing legacy-only assertions still hold, and the gRPC contract test still passes (current instrumentation emits the legacy key, which flows through the fallback).

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/SpanMetricsProcessor.java
```diff
@@ -0,0 +1,176 @@
...
+      if (openTelemetry == null) {
+        return false;
+      }
+      Meter meter = openTelemetry.meterBuilder(SCOPE_NAME).build();
+      // No metrics pipeline (e.g. OTEL_METRICS_EXPORTER=none or a reader-less MeterProvider): the
+      // SDK hands back the API no-op meter. Recording would go nowhere, yet stamping the dedup
+      // marker would still make the backend skip generation — so no side produces the metric. Stay
+      // inert instead, so the backend generates from the sampled spans and the exactly-once
+      // invariant
+      // holds. Not cached: binding is late, so a real pipeline may appear on a later span.
```
liustve, line 154: non-blocking: Once a bound reader-less provider reaches this branch, its reader set cannot change and first-bind-wins prevents replacing it, but `calls` remains null so every span start and end re-enters the synchronized initializer. A concurrent probe retained 43.5-52% of baseline throughput at 4-12 threads, so consider caching a terminal disabled state after a non-null holder yields this no-op meter while continuing to retry only before binding.

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/SpanMetricsProcessor.java
```diff
@@ -0,0 +1,176 @@
...
+    // metrics were already generated by the extension (do not regenerate). Only stamp once we know
+    // the host actually has a metrics pipeline that can deliver our metrics — otherwise the marker
+    // would suppress backend generation for a metric the client never emits (see
+    // ensureInstruments).
+    // Guarded so a failure here can never propagate into the host application's span pipeline.
+    try {
+      if (!ensureInstruments()) {
+        return;
+      }
+      span.setAttribute(SCHEMA_ATTR, SCHEMA_VERSION);
```
liustve, line 99: nit (non-blocking): These load-bearing dedup attributes consume ordinary bounded span-attribute slots. With a valid max-attribute limit of `1` already filled before `onStart`, a real SDK probe recorded the client metric but exported no schema marker, allowing backend duplication; consider defining the exhausted-budget behavior and adding a low-positive-limit regression.

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/SpanMetricsProcessor.java
```diff
@@ -0,0 +1,176 @@
...
+      if (!ensureInstruments()) {
+        return;
+      }
+      // One toSpanData() snapshot per span: status and resource are only reachable via SpanData on
+      // the ReadableSpan interface, so this allocation is inherent to the contract (reading fields
+      // off the concrete SdkSpan would break non-SDK ReadableSpan implementations). Everything else
+      // is read from this single snapshot.
+      SpanData data = span.toSpanData();
+      Attributes attributes = SpanMetricsAttributesBuilder.build(data);
+      calls.add(1, attributes);
```
liustve, line 123: nit (non-blocking): This two-argument recording call resolves `Context.current()`, so ending a child after closing its scope can attach the still-current parent's exemplar to the child's datapoint even though the v191 client contract says exemplars are unsupported in v1. Consider choosing an explicit context policy for both instruments and covering a nested-span case.

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/SpanMetricsProcessor.java
```diff
@@ -0,0 +1,176 @@
...
+  @Override
+  public void onStart(Context parentContext, ReadWriteSpan span) {
+    // Dedup + schema/lib markers: presence of the schema attribute tells the backend this span's
+    // metrics were already generated by the extension (do not regenerate). Only stamp once we know
+    // the host actually has a metrics pipeline that can deliver our metrics — otherwise the marker
+    // would suppress backend generation for a metric the client never emits (see
+    // ensureInstruments).
+    // Guarded so a failure here can never propagate into the host application's span pipeline.
+    try {
+      if (!ensureInstruments()) {
```
liustve, line 96: nit (non-blocking): Evaluating readiness independently at start and end creates a narrow bind-transition race: a span started before manual `bind()` receives no marker, but if it ends after binding the client metric is recorded, so an exported sampled span can also be metered by the backend. Consider adding a start-before-bind/end-after-bind regression and documenting the intended transition behavior.

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/SpanMetricsProcessor.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+    return true;
+  }
+
+  // A reader-less SdkMeterProvider returns the API's no-op meter — the same class
+  // MeterProvider.noop()
+  // yields. Comparing against that stays correct across SDK versions (the concrete no-op class name
+  // has changed between releases), unlike matching a hard-coded class name.
+  private static boolean isNoopMeter(Meter meter) {
+    return meter.getClass().equals(NOOP_METER_CLASS);
```
liustve, line 174: nit (non-blocking): A non-no-op meter class does not prove these instruments survive host views or reader filtering. With a real registered reader plus an `Aggregation.drop()` view for this scope, probes on SDK 1.32.0, 1.44.1, and 1.65.0 produced marked spans but zero metrics; consider documenting filtered pipelines as unsupported or adding a registered-reader/drop-view policy test.

cloudwatch-plugin-otel/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+### 2. Spring Boot starter
+
+Put the jar on the classpath of a Spring Boot 3 app that uses the OpenTelemetry Spring Boot starter.
+It is auto-configured — no code changes.
+
+### 3. Plain SDK autoconfigure
+
+Put the jar on the classpath of an app using `opentelemetry-sdk-extension-autoconfigure`. The
+plugin's `AutoConfigurationCustomizerProvider` is discovered automatically — no code changes.
```
liustve, line 78: nit (non-blocking): This zero-code autoconfigure claim does not hold when `otel.experimental.config.file` is used because upstream returns the file-built SDK before applying tracer/sampler customizers or post-build listeners, leaving the plugin inert. Since the plugin has no hook into that path, consider documenting declarative file configuration as unsupported.

cloudwatch-plugin-otel/README.md
```diff
@@ -0,0 +1,124 @@
...
+Span metrics are the first feature of this plugin; more CloudWatch-oriented OpenTelemetry features
+may be added over time.
+
+## What it produces
+
+Two metrics, matching the OpenTelemetry SpanMetrics naming:
+
+| Metric | Instrument | Unit |
+| --- | --- | --- |
+| `traces.span.metrics.calls` | Counter (monotonic sum) | `1` |
```
liustve, line 21: nit (non-blocking): The counter builder leaves its unit empty, the unit test asserts an empty unit, and specification v191 requires it to be unset for connector parity. Please change this cell from `1` to `unset` rather than changing the emitted descriptor.

cloudwatch-plugin-otel/README.md
```diff
@@ -0,0 +1,124 @@
...
+| `traces.span.metrics.duration` | Histogram | `s` (seconds) |
+
+Each datapoint carries low-cardinality dimensions: `service.name`, `span.name`, `span.kind`,
```
liustve, line 24: nit (non-blocking): This output summary omits `aws.otel.span.metrics.schema` and `aws.otel.extension.lib.version`, and it does not disclose that the processor also adds both attributes to every recorded application span. Consider documenting both metric dimensions and the observable span mutation.

liustve:
nit (non-blocking): The PR description still says this was opened as a draft, names the obsolete `otel.cloudwatch.spanmetrics` scope instead of `cloudwatch.plugin.otel.span_metrics`, and lists remaining work that no longer matches the current state. Please refresh the description before merge.

liustve (approved):
PTAL at some of the not blocking comments

wangzlei (approved):

cloudwatch-plugin-otel/README.md
srprash: Thanks — agreed. This is a docs-only change; batching it into the follow-up README/docs PR after this one merges (noted in the PR description). Resolving to keep the merge queue clear.

cloudwatch-plugin-otel/README.md
srprash: Thanks — agreed. This is a docs-only change; batching it into the follow-up README/docs PR after this one merges (noted in the PR description). Resolving to keep the merge queue clear.

cloudwatch-plugin-otel/README.md
srprash: Thanks — agreed. This is a docs-only change; batching it into the follow-up README/docs PR after this one merges (noted in the PR description). Resolving to keep the merge queue clear.

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/SpanMetricsProcessor.java
srprash: Thanks — agreed this is non-blocking. Tracking it to address in a follow-up PR rather than expanding this one's scope. Resolving for now.

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/SpanMetricsProcessor.java
srprash: Thanks — agreed this is non-blocking. Tracking it to address in a follow-up PR rather than expanding this one's scope. Resolving for now.

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/SpanMetricsProcessor.java
srprash: Thanks — agreed this is non-blocking. Tracking it to address in a follow-up PR rather than expanding this one's scope. Resolving for now.

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/SpanMetricsProcessor.java
srprash: Thanks — agreed this is non-blocking. Tracking it to address in a follow-up PR rather than expanding this one's scope. Resolving for now.

cloudwatch-plugin-otel/src/main/java/software/amazon/opentelemetry/cloudwatch/spanmetrics/SpanMetricsProcessor.java
srprash: Thanks — agreed this is non-blocking. Tracking it to address in a follow-up PR rather than expanding this one's scope. Resolving for now.

PR #1439, opened 2026-08-18
fix: replace SHA-1 and MD5 with SHA-256 for FIPS compliance

amdonov:
The serviceevents module uses SHA-1 (EndpointIdGenerator) and MD5 (IncidentRateLimiter) for deterministic ID generation. These algorithms are prohibited under FIPS 140-3 as of December 2025 (NIST SHA-1 retirement), causing the agent to fail in environments that enforce strict FIPS security policies blocking legacy algorithms.

Changes:
- EndpointIdGenerator: Replace SHA-1 UUID5 with SHA-256-based name UUID using version nibble 8 (custom, per RFC 9562).
- IncidentRateLimiter: Replace MD5 with SHA-256 (truncated to 128 bits) for dedup hash keys, preserving output length compatibility.

Both usages are non-cryptographic (deterministic ID generation only). Generated IDs will differ from prior versions.

Tested with ACCP-FIPS (AWS-LC FIPS certificate #4816) under a strict security policy that disables SHA-1 and MD5 via
jdk.tls.disabledAlgorithms and jdk.certpath.disabledAlgorithms. Agent starts and exports telemetry successfully.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ If your change does not need a CHANGELOG entry, add the "skip changelog" label t
 
 ## Unreleased
 
+- fix(serviceevents): replace SHA-1 and MD5 with SHA-256 for FIPS 140-3 compliance
+  ([#1439](https://github.com/aws-observability/aws-otel-java-instrumentation/pull/1439))
 - feat: attribute presigned S3 URLs as `AWS::S3` dependencies in Application Signals, opt-in via
   `otel.aws.application.signals.presigned-url-attribution.enabled`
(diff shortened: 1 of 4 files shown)


amdonov:
I'm converting this to draft. While a customer reported errors related to SHA-1 use in a FIPS 140-3 environment, I believe this is a result of an excessive security policy blocking all SHA-1 use. NIST allows the use of SHA-1 for "Non-digital-signature applications". Will follow up once I have more details.

PR #688, opened 2024-01-02
Fix Typo in gh action 

vasireddy99:
*Description of changes:*

Fix Typo in gh action 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1447, opened 2026-09-02
Java eks test fix

haneric00:
(in application-signals-e2e-test.yml)

Orchestrates the shared-cluster EKS leg:
- eks-setup installs App Signals once → eks-patch-java-operator sets the operator's Java image once → the version jobs + service-events-eks run in parallel → eks-cleanup tears App Signals down once, gated on all of them.
- Each eks-v{8,11,17,21,25}-amd64 and service-events-eks calls the parallel reusable workflows and depends on the patch gate.
- Concurrency cap: eks-v25-amd64 depends on eks-v8-amd64 (ordering only), so at most 4 versions' pod-sets are ever on the cluster at once. Empirically 5 concurrent exceeded node capacity and one version's pods failed to become Ready; 4 fit.
- Updated the validate-e2e-tests-are-accounted-for exclusions to include the new gate workflow (it's not a *-test.yml suite).

Why the cap

With all 5 running, exactly one version failed per run — a different one each time (v11, then v17, then v21). That "rotating single loser" is the signature of just-insufficient cluster capacity, not a version bug. Capping peak at 4 keeps the run within the cluster's headroom while preserving most of the parallelism win.


.github/workflows/application-signals-e2e-test.yml
```diff
@@ -315,3 +346,5 @@ jobs:
   #   (see: https://github.com/aws-observability/aws-application-signals-test-framework/blob/main/.github/workflows/validate-e2e-tests-are-accounted-for.yml)
   validate-all-tests-are-accounted-for:
     uses: aws-observability/aws-application-signals-test-framework/.github/workflows/validate-e2e-tests-are-accounted-for.yml@main
+    with:
+      exclusions: 'java-eks-test.yml,java-eks-service-events-test.yml,eks-setup-app-signals.yml,eks-cleanup-app-signals.yml,java-eks-patch-java-operator.yml'
```
jj22ee: **Blocker — and it's already broken.**

`aws/amazon-cloudwatch-agent`'s `validate-all-tests-are-accounted-for` job is failing right now: [run 33768518178](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/33768518178) (2026-09-03), and every E2E run since 2026-08-28.

Root cause is the same issue raised on #852 that didn't get actioned: PR #682 added `python-eks-parallel-test.yml` to the framework, that repo's `TESTS_EXPECTED` pattern is `(^java-)|(^node-)|(^python-)|(^dotnet-)|(^metric-limiter-)` filtered by `test\.ya?ml$`, and its exclusions are still just `dotnet-ec2-nuget-test.yml`.

Framework PR #688 adds two more files matching that pattern (`java-eks-parallel-test.yml`, `java-eks-service-events-parallel-test.yml`), so it compounds a break that's already red. Could we land a PR against `aws/amazon-cloudwatch-agent` adding all three to `exclusions` **before** #688 merges?

(`amazon-cloudwatch-agent-operator` also calls `java-eks-test.yml` but doesn't call the validator, so it's unaffected here.)

Separately, three entries in this list are no-ops: `eks-setup-app-signals.yml`, `eks-cleanup-app-signals.yml`, and `java-eks-patch-java-operator.yml` don't match `test\.ya?ml$`, so they were never in `TESTS_EXPECTED`. Harmless, but misleading to the next person — I'd drop them and keep just the two sequential workflows.

.github/workflows/application-signals-e2e-test.yml
```diff
@@ -130,9 +130,32 @@ jobs:
   # DEFAULT SETTING: {Java Version}, EKS, AMD64, AL2
   #
 
-  eks-v8-amd64:
-    uses: aws-observability/aws-application-signals-test-framework/.github/workflows/java-eks-test.yml@main
+  # --- Parallel Java EKS pipeline -------------------------------------------------------------
+  # Shape: metric-limiter -> eks-setup -> [eks-v8..v25 + service-events-eks concurrently] -> eks-cleanup
+
+  eks-setup:
+    needs: [ metric-limiter-v11-amd64 ]
```
jj22ee: This puts `metric-limiter` serially in front of the entire EKS band, so the critical path is now `metric-limiter + eks-setup + patch + slowest version + eks-cleanup`.

Could you post the measured before/after wall-clock? The "almost an hour" figure is the whole justification for ~1,000 lines across two repos, and with setup/patch/cleanup added serially at both ends I'd like to see where it actually nets out.

.github/workflows/application-signals-e2e-test.yml
```diff
@@ -130,9 +130,32 @@ jobs:
...
+    if: ${{ always() }}
```
jj22ee: `always()` here isn't just redundant — it makes this job launch **even when the run is cancelled**, so cancelling a bad E2E run would still install App Signals on the shared cluster and let the downstream terraform applies fire.

If the intent is "run the EKS band even if metric-limiter failed," `!cancelled()` expresses that without the cancellation hole. If the intent is just ordering, `needs:` alone is enough.

.github/workflows/application-signals-e2e-test.yml
```diff
@@ -130,9 +130,32 @@ jobs:
...
+    uses: aws-observability/aws-application-signals-test-framework/.github/workflows/eks-setup-app-signals.yml@main
     secrets: inherit
+    with:
+      test-cluster-name: 'e2e-adot-test'
+      aws-region: us-east-1
+
+  eks-patch-java-operator:
+    needs: eks-setup
+    if: ${{ always() && needs.eks-setup.result == 'success' }}
```
jj22ee: Worth noting the re-run trap is a bit worse than in the Python version: re-running one failed version leg now skips **both** `eks-setup` and this patch gate, so the leg burns all 5 inject-verification attempts against an unpatched (or torn-down) cluster before failing.

If that's accepted as the design for now, can the failure message in `java-eks-parallel-test.yml` say so explicitly — something like "re-run the entire workflow, not just failed jobs"? Otherwise the first person to hit this at 5pm loses a while to it.

.github/workflows/application-signals-e2e-test.yml
```diff
@@ -130,9 +130,32 @@ jobs:
...
+    uses: aws-observability/aws-application-signals-test-framework/.github/workflows/java-eks-patch-java-operator.yml@main
+    secrets: inherit
+    with:
+      test-cluster-name: 'e2e-adot-test'
+      aws-region: us-east-1
+      adot-image-name: ${{ inputs.adot-image-name }}
+
+  eks-v8-amd64:
+    needs: [ eks-setup, eks-patch-java-operator ]
+    if: ${{ always() && needs.eks-patch-java-operator.result == 'success' }}
```
jj22ee: This `if:` can be deleted entirely — `needs: [eks-setup, eks-patch-java-operator]` already means "only run if both succeeded," and dropping `always()` restores correct cancellation behaviour. Same for `eks-v11`/`v17`/`v21` and `service-events-eks`.

(`eks-v25` is the one exception — see my comment there.)

.github/workflows/application-signals-e2e-test.yml
```diff
@@ -174,16 +200,25 @@ jobs:
       java-version: '21'
 
   eks-v25-amd64:
-    needs: eks-v21-amd64
-    uses: aws-observability/aws-application-signals-test-framework/.github/workflows/java-eks-test.yml@main
-    secrets: inherit
+    needs: [ eks-setup, eks-patch-java-operator, eks-v8-amd64 ]
```
jj22ee: This is the one place `always()` is actually load-bearing — you want `eks-v8` for ordering only, without failure propagation — so I don't want it removed blindly. But two things:

1. `!cancelled()` is the right idiom rather than `always()`, so a cancelled run doesn't still fire this.
2. **The reason for `needs: eks-v8-amd64` exists only in the PR description.** Nothing in the YAML says "this dependency is a concurrency cap, not a real dependency." Someone will tidy it away and silently reintroduce the rotating-loser capacity failures.

GitHub has the built-in for this — a matrix with `max-parallel`:

```yaml
  eks-version-test:
    needs: [ eks-setup, eks-patch-java-operator ]
    strategy:
      fail-fast: false
      max-parallel: 4   # 5 versions' pod-sets exceed node capacity; 4 fit
      matrix:
        java-version: [ '8', '11', '17', '21', '25' ]
    uses: aws-observability/aws-application-signals-test-framework/.github/workflows/java-eks-parallel-test.yml@main
    secrets: inherit
    with:
      aws-region: us-east-1
      test-cluster-name: 'e2e-adot-test'
      adot-image-name: ${{ inputs.adot-image-name }}
      caller-workflow-name: 'main-build'
      java-version: ${{ matrix.java-version }}
```

`strategy` is allowed on `uses:` jobs, and `java-lambda-test` in this same file already uses a matrix. This collapses five near-identical blocks into one, makes the cap declarative and self-documenting, and gives you one place to manage the supported version list — which is also what the `java_versions` terraform comment in #688 is asking for.

.github/workflows/application-signals-e2e-test.yml
```diff
@@ -174,16 +200,25 @@ jobs:
...
+    uses: aws-observability/aws-application-signals-test-framework/.github/workflows/java-eks-parallel-test.yml@main
     with:
       aws-region: us-east-1
       test-cluster-name: 'e2e-adot-test'
       adot-image-name: ${{ inputs.adot-image-name }}
       caller-workflow-name: 'main-build'
       java-version: '25'
 
+  eks-cleanup:
+    if: ${{ always() && needs.eks-setup.result == 'success' }}
```
jj22ee: This gate has the same bug as the merged Python version. `needs.eks-setup.result == 'success'` means a setup run that fails **after** installing the addon (e.g. the final IRSA verification) skips cleanup entirely, leaving the shared cluster fully configured with nothing to reap it.

Should be:

```yaml
    if: ${{ always() && needs.eks-setup.result != 'skipped' }}
```

`always()` genuinely is wanted here (you want teardown after failures), it's just the gate that's wrong.

.github/workflows/application-signals-e2e-test.yml
jj22ee: **Blocker — and it's already broken.**

`aws/amazon-cloudwatch-agent`'s `validate-all-tests-are-accounted-for` job is failing right now: [run 33768518178](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/33768518178) (2026-09-03), and every E2E run since 2026-08-28.

Root cause is the same issue raised on #852 that didn't get actioned: PR #682 added `python-eks-parallel-test.yml` to the framework, that repo's `TESTS_EXPECTED` pattern is `(^java-)|(^node-)|(^python-)|(^dotnet-)|(^metric-limiter-)` filtered by `test\.ya?ml$`, and its exclusions are still just `dotnet-ec2-nuget-test.yml`.

Framework PR #688 adds two more files matching that pattern (`java-eks-parallel-test.yml`, `java-eks-service-events-parallel-test.yml`), so it compounds a break that's already red. Could we land a PR against `aws/amazon-cloudwatch-agent` adding all three to `exclusions` **before** #688 merges?

(`amazon-cloudwatch-agent-operator` also calls `java-eks-test.yml` but doesn't call the validator, so it's unaffected here.)

.github/workflows/application-signals-e2e-test.yml
jj22ee: Re-running a single failed version leg skips both `eks-setup` and this patch gate, so the leg burns all 5 inject-verification attempts against an unpatched cluster before failing. I know that tradeoff was already settled in the Python round — could the failure message in `java-eks-parallel-test.yml` just say so explicitly, something like "re-run the entire workflow, not just failed jobs"? Saves the next person the debugging trip.

.github/workflows/application-signals-e2e-test.yml
jj22ee: This `if:` can be deleted entirely — `needs: [eks-setup, eks-patch-java-operator]` already means "only run if both succeeded," and dropping `always()` restores correct cancellation behaviour. As written, cancelling a bad E2E run still fires terraform applies against the shared cluster. Same for `eks-v11`/`v17`/`v21` and `service-events-eks`.

Two exceptions worth calling out: `eks-v25` genuinely needs the override (see my comment there), and on `eks-setup` (line 138) `!cancelled()` is probably what you want rather than `always()` — running the EKS band despite a metric-limiter failure looks intentional, but launching it on a cancelled run isn't.

.github/workflows/application-signals-e2e-test.yml
jj22ee: Only two of these are needed: `java-eks-test.yml` and `java-eks-service-events-test.yml`. The other three never enter `TESTS_EXPECTED` anyway — `eks-setup-app-signals.yml` and `eks-cleanup-app-signals.yml` don't match the `^java-`/`^metric-limiter-` prefix filter, and `java-eks-patch-java-operator.yml` doesn't end in `test.yml`.

.github/workflows/application-signals-e2e-test.yml
jj22ee: This `if:` can be deleted entirely, and the merged Python caller is the precedent — its `eks-py3XX` jobs are just `needs: eks-setup` with no `if:` at all. `needs:` already means "only run if it succeeded," and dropping `always()` restores correct cancellation behaviour; as written, cancelling a bad E2E run still fires terraform applies against the shared cluster. Same for `eks-v11`/`v17`/`v21` and `service-events-eks`.

Two exceptions: `eks-v25` genuinely needs an override (see my comment there), and on `eks-setup` (line 138) the merged Python caller uses `if: ${{ !cancelled() }}` — which is what you want, since running the EKS band despite a metric-limiter failure looks intentional but launching it on a cancelled run isn't.

.github/workflows/application-signals-e2e-test.yml
jj22ee: This gate is wrong, and I think I can point at how it got here. The merged Python caller's `eks-cleanup` is just `if: ${{ always() }}` with no result gate — but the comment above it still reads "gated on eks-setup success so it never tries to clean up state that was never installed," describing a gate that was removed before merge. This PR implements the stale comment rather than the code.

The consequence is a leak: `needs.eks-setup.result == 'success'` means a setup run that fails *after* installing the addon (e.g. the final IRSA verification) skips cleanup entirely, leaving the shared cluster configured with nothing to reap it. Either match Python's plain `if: ${{ always() }}`, or use `!= 'skipped'` if you want to skip a no-op cleanup when setup never ran. Worth fixing Python's comment while you're at it.

.github/workflows/application-signals-e2e-test.yml
jj22ee: The merged Python workflow already solved this — `python-eks-shared-cluster-test.yml` ends with a `How to re-run this test` step (`if: failure()`) that prints:

```
  This test job cannot be re-run on its own.
  => Re-run the pipeline starting from eks-setup so it flows:
       eks-setup -> eks-py3xx (this test) -> eks-cleanup
```

Can we port that into `java-eks-parallel-test.yml`? It matters more here than in Python, since a lone re-run skips both `eks-setup` and this patch gate, so the leg burns all 5 inject-verification attempts before failing.

.github/workflows/application-signals-e2e-test.yml
jj22ee, line 350: Only these two need excluding:

```suggestion
      exclusions: 'java-eks-test.yml,java-eks-service-events-test.yml'
```

jj22ee: **Blocker — and it's already broken.**

`aws/amazon-cloudwatch-agent`'s `validate-all-tests-are-accounted-for` job is failing right now: [run 33768518178](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/33768518178) (2026-09-03), and every E2E run since 2026-08-28. PR #682 added `python-eks-parallel-test.yml` to the framework, that repo's expected-test filter includes `^python-` + `test\.ya?ml$`, and its exclusions are still just `dotnet-ec2-nuget-test.yml`.

Framework PR #688 adds two more files matching those filters, so it compounds a break that's already red. Can a PR against `aws/amazon-cloudwatch-agent` — adding `python-eks-parallel-test.yml`, `java-eks-parallel-test.yml`, and `java-eks-service-events-parallel-test.yml` to its existing `exclusions` — land before #688 merges?

.github/workflows/application-signals-e2e-test.yml
jj22ee: Two edge cases as written: `always()` starts `eks-v25` even on a cancelled run, so cancelling mid-band kicks off another terraform apply against the shared cluster. And `needs: eks-v8-amd64` reads as a real dependency — nothing marks it as a concurrency cap, so it's easy to tidy away and silently hit the capacity failures again.

A matrix handles both:

```yaml
  eks-version-test:
    needs: [ eks-setup, eks-patch-java-operator ]
    strategy:
      fail-fast: false
      max-parallel: 4   # 5 versions' pod-sets exceed node capacity; 4 fit
      matrix:
        java-version: [ '8', '11', '17', '21', '25' ]
    uses: aws-observability/aws-application-signals-test-framework/.github/workflows/java-eks-parallel-test.yml@main
    secrets: inherit
    with:
      aws-region: us-east-1
      test-cluster-name: 'e2e-adot-test'
      adot-image-name: ${{ inputs.adot-image-name }}
      caller-workflow-name: 'main-build'
      java-version: ${{ matrix.java-version }}
```

.github/workflows/application-signals-e2e-test.yml
jj22ee, line 157: `if` + `always()` here is already covered by `needs` — and `always()` also makes this run on cancelled runs. So we should be able to remove the `if:` here. Same for `eks-v11`/`v17`/`v21` and `service-events-eks`.

.github/workflows/application-signals-e2e-test.yml
jj22ee, line 214: `== 'success'` leaks the cluster: if setup fails *after* installing the addon (e.g. the final IRSA check), cleanup is skipped and the shared cluster stays configured. `!= 'skipped'` covers it — the merged Python caller just uses a [bare `always()`](https://github.com/aws-observability/aws-otel-python-instrumentation/blob/main/.github/workflows/application-signals-e2e-test.yml#L207-L209).

.github/workflows/application-signals-e2e-test.yml
jj22ee, line 147: Re-running a single failed version job skips both `eks-setup` and this gate, so it burns all 5 inject-verification attempts against a torn-down cluster before failing. Python prints guidance for exactly this — could you port its [`How to re-run this test`](https://github.com/aws-observability/aws-application-signals-test-framework/blob/main/.github/workflows/python-eks-shared-cluster-test.yml#L472-L485) step (`if: failure()`) into `java-eks-parallel-test.yml`?

.github/workflows/application-signals-e2e-test.yml
```diff
@@ -174,16 +200,25 @@ jobs:
       java-version: '21'
 
   eks-v25-amd64:
-    needs: eks-v21-amd64
-    uses: aws-observability/aws-application-signals-test-framework/.github/workflows/java-eks-test.yml@main
-    secrets: inherit
+    needs: [ eks-setup, eks-patch-java-operator, eks-v8-amd64 ]
+    if: ${{ always() && needs.eks-patch-java-operator.result == 'success' }}
```
jj22ee: Two edge cases as written: `always()` starts this even on a cancelled run, and `needs: eks-v8-amd64` reads as a real dependency rather than a concurrency cap, so it's easy to remove and silently hit the capacity failures again.

```suggestion
    # eks-v8-amd64 is ordering-only: it caps peak concurrency at 4 version jobs,
    # since 5 concurrent pod-sets exceed node capacity.
    needs: [ eks-setup, eks-patch-java-operator, eks-v8-amd64 ]
    if: ${{ !cancelled() && needs.eks-patch-java-operator.result == 'success' }}
```

If you'd rather do it declaratively, a `strategy: matrix` with `max-parallel: 4` replaces all five version job blocks and makes the cap self-documenting — but the above is the smaller change.

.github/workflows/application-signals-e2e-test.yml
jj22ee, line 203: `needs: eks-v8-amd64` reads as a real dependency rather than a concurrency cap, so it's easy to tidy away and silently hit the capacity failures again. Could you add the context inline?

```suggestion
    # eks-v8-amd64 is ordering-only: it caps peak concurrency at 4 version jobs,
    # since 5 concurrent pod-sets exceed node capacity.
    needs: [ eks-setup, eks-patch-java-operator, eks-v8-amd64 ]
```

jj22ee (commented):
Added comments with help from AI

.github/workflows/application-signals-e2e-test.yml
```diff
@@ -130,59 +130,99 @@ jobs:
...
+    needs: [ eks-setup, eks-patch-java-operator ]
+    uses: aws-observability/aws-application-signals-test-framework/.github/workflows/java-eks-shared-cluster-test.yml@java-eks-test-fix
     secrets: inherit
     with:
       aws-region: us-east-1
       test-cluster-name: 'e2e-adot-test'
       adot-image-name: ${{ inputs.adot-image-name }}
       caller-workflow-name: 'main-build'
       java-version: '8'
+      java-versions: '["8","11","17","21","25"]'
```
jj22ee: This list is repeated in all six EKS jobs, so adding a version means editing six places which is prone to error. Can you centralize/hoist it to a non-required `workflow_call` input, the way [`aws_region`](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/workflows/release-build.yml#L8) works in `release-build.yml`?

```yaml
on:
  workflow_call:
    inputs:
      java-versions:
        required: false
        type: string
        default: '["8","11","17","21","25"]'
```

jj22ee (approved):

syed-ahsan-ishtiaque (approved):

